PR #2, opened 2023-04-05
initial implementation of transport layer

johncantrell97:
As title says, just a first pass on implementing the LSPS transport layer.  Looking to get feedback on this direction.

I thought we would want to expose the raw jsonrpc::Request/Response/Notification interface to library users because each LSP will likely end up supporting different subset of the full spec and probably custom extensions.  This is why I added the `LSPMessageHandler` trait that lets someone implement handlers for it.  I assume when we implement LSPS request-based protocols we'd provide a handler that handles the logic for that particular protocol.

I'm not so sure on this `LSPManager` object.  My thought was it handled the layer of going from raw string <==> rpc request/response/notification for the user.  It would also expose some api to a higher level and perform whatever bookkeeping is necessary (managing pending request's / ids for them?).

I could use some ideas for how to proceed from here.  I probably should think about how a user of the library will want to use it and work backwards from there.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,6 @@ lightning-invoice = { version = "0.22" }
 lightning-net-tokio = { version = "0.0.114" }
 
 bitcoin = "0.29.2"
+
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
--- a/src/events.rs
+++ b/src/events.rs
@@ -1,0 +1,10 @@
+use bitcoin::secp256k1::PublicKey;
+
+pub trait EventHandler {
+	fn handle_event(&self, event: Event);
+}
+
+#[derive(Debug, Clone)]
+pub enum Event {
+	ListProtocols { counterparty_node_id: PublicKey, protocols: Vec<u16> },
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,12 +11,15 @@
 //! # ldk-lsp-client
 //! Types and primitives to integrate a spec-compliant LSP with an LDK-based node.
 #![deny(missing_docs)]
-#![deny(broken_intra_doc_links)]
-#![deny(private_intra_doc_links)]
+#![deny(rustdoc::broken_intra_doc_links)]
+#![deny(rustdoc::private_intra_doc_links)]
 #![allow(bare_trait_objects)]
 #![allow(ellipsis_inclusive_range_patterns)]
+#![allow(clippy::drop_non_drop)]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
 mod channel_request;
+mod events;
 mod jit_channel;
 mod transport;
+mod utils;
--- a/src/transport/jsonrpc.rs
+++ b/src/transport/jsonrpc.rs
@@ -1,0 +1,270 @@
+// This file is Copyright its original authors, visible in version contror
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! Types and primitives for [JSON-RPC 2.0 specification](https://www.jsonrpc.org/specification).
+use std::{collections::HashMap, fmt::Display, str::FromStr};
+
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+pub const VERSION: &str = "2.0";
+pub const INVALID_REQUEST_ERROR_CODE: i32 = -32600;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Method {
+	pub prefix: String,
+	pub name: String,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct ParseMethodError;
+
+impl Display for ParseMethodError {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		write!(f, "invalid method format")
+	}
+}
+
+impl FromStr for Method {
+	type Err = ParseMethodError;
+
+	fn from_str(s: &str) -> Result<Self, Self::Err> {
+		let mut split = s.split('.');
+		let prefix = split.next().ok_or(ParseMethodError)?.to_string();
+		let name = split.next().ok_or(ParseMethodError)?.to_string();
+		Ok(Method { prefix, name })
+	}
+}
+
+impl Display for Method {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		write!(f, "{}.{}", self.prefix, self.name)
+	}
+}
+
+impl Serialize for Method {
+	fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+	where
+		S: serde::Serializer,
+	{
+		serializer.serialize_str(&self.to_string())
+	}
+}
+
+// implement Deserialize for Method using it's FromStr implementation
+impl<'de> Deserialize<'de> for Method {
+	fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+	where
+		D: serde::Deserializer<'de>,
+	{
+		let s = String::deserialize(deserializer)?;
+		Method::from_str(&s).map_err(serde::de::Error::custom)
+	}
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Error {
+	/// A number indicating the error type that occurred.
+	pub code: i32,
+	/// A string providing a short description of the error.
+	pub message: String,
+	/// A primitive or structured value that contains additional information about the error.
+	pub data: Option<Value>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+/// A JSONRPC request object.
+pub struct Request {
+	/// The name of the RPC call.
+	pub method: Method,
+	/// Parameters to the RPC call.
+	/// Note: Only by-name parameters are allowed.
+	pub params: HashMap<String, Value>,
+	/// Identifier for this Request, which should appear in the response.
+	pub id: Value,
+	/// jsonrpc field, MUST be "2.0".
+	pub jsonrpc: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Notification {
+	/// The name of the RPC call.
+	pub method: Method,
+	/// Parameters to the RPC call.
+	pub params: HashMap<String, Value>,
+	/// jsonrpc field, MUST be "2.0".
+	pub jsonrpc: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(untagged)]
+/// A JSON-RPC 2.0 response object.
+pub enum Response {
+	/// A successful response.
+	Success(SuccessResponse),
+	/// An error response.
+	Error(ErrorResponse),
+}
+
+impl Response {
+	pub fn get_id(&self) -> &Value {
+		match self {
+			Response::Success(success) => &success.id,
+			Response::Error(error) => &error.id,
+		}
+	}
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+/// A JSON-RPC 2.0 error response object.
+pub struct ErrorResponse {
+	/// The error.
+	pub error: Error,
+	/// Identifier for this Request, which should match that of the request.
+	pub id: Value,
+	/// jsonrpc field, MUST be "2.0".
+	pub jsonrpc: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+/// A JSON-RPC 2.0 successful response object.
+pub struct SuccessResponse {
+	/// The result.
+	pub result: Value,
+	/// Identifier for this Request, which should match that of the request.
+	pub id: Value,
+	/// jsonrpc field, MUST be "2.0".
+	pub jsonrpc: String,
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use serde_json::json;
+
+	#[test]
+	fn deserializes_error_response() {
+		let json = r#"{
+            "jsonrpc": "2.0",
+            "error": {
+                "code": -32600,
+                "message": "Invalid Request"
+            },
+            "id": null
+        }"#;
+
+		let response: Response = serde_json::from_str(json).unwrap();
+		assert_eq!(
+			response,
+			Response::Error(ErrorResponse {
+				error: Error { code: -32600, message: "Invalid Request".to_string(), data: None },
+				id: serde_json::Value::Null,
+				jsonrpc: "2.0".to_string(),
+			})
+		);
+	}
+
+	#[test]
+	fn deserializes_success_response() {
+		let json = r#"{
+            "jsonrpc": "2.0",
+            "result": 5,
+            "id": "1"
+        }"#;
+
+		let response: Response = serde_json::from_str(json).unwrap();
+		assert_eq!(
+			response,
+			Response::Success(SuccessResponse {
+				result: json!(5),
+				id: json!("1"),
+				jsonrpc: "2.0".to_string(),
+			})
+		);
+	}
+
+	#[test]
+	fn fails_deserialize_with_both_error_result() {
+		let json = r#"{
+            "jsonrpc": "2.0",
+            "result": 5,
+            "error": {
+                "code": -32600,
+                "message": "Invalid Request"
+            },
+            "id": "1"
+        }"#;
+
+		let response: Result<Response, serde_json::Error> = serde_json::from_str(json);
+		assert!(response.is_err());
+	}
+
+	#[test]
+	fn serializes_error_response() {
+		let response = Response::Error(ErrorResponse {
+			error: Error { code: -32600, message: "Invalid Request".to_string(), data: None },
+			id: serde_json::Value::Null,
+			jsonrpc: "2.0".to_string(),
+		});
+
+		let json = serde_json::to_string(&response).unwrap();
+		assert_eq!(
+			json,
+			r#"{"error":{"code":-32600,"message":"Invalid Request","data":null},"id":null,"jsonrpc":"2.0"}"#
+		);
+	}
+
+	#[test]
+	fn serializes_success_response() {
+		let response = Response::Success(SuccessResponse {
+			result: json!(5),
+			id: json!("1"),
+			jsonrpc: "2.0".to_string(),
+		});
+
+		let json = serde_json::to_string(&response).unwrap();
+		assert_eq!(json, r#"{"result":5,"id":"1","jsonrpc":"2.0"}"#);
+	}
+
+	#[test]
+	fn deserializes_request_method() {
+		let json = r#"{
+			"jsonrpc": "2.0",
+			"method": "prefix.name",
+			"params": { "a": 5 },
+			"id": 3
+		}"#;
+
+		let request: Request = serde_json::from_str(json).unwrap();
+		assert_eq!(
+			request,
+			Request {
+				method: Method { prefix: "prefix".to_string(), name: "name".to_string() },
+				params: vec![("a".to_string(), json!(5)),].into_iter().collect(),
+				id: json!(3),
+				jsonrpc: "2.0".to_string(),
+			}
+		);
+	}
+
+	#[test]
+	fn serializes_request_method() {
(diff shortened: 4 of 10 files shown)


src/transport/message_handler.rs
```diff
@@ -0,0 +1,74 @@
...
+impl CustomMessageReader for LSPSMessageHandler {
+	type CustomMessage = msgs::LSPSMessage;
+
+	fn read<R: io::Read>(
+		&self, message_type: u16, buffer: &mut R,
+	) -> Result<Option<Self::CustomMessage>, lightning::ln::msgs::DecodeError> {
+		match message_type {
+			37913 => {
+				let msg: msgs::LSPSMessage = serde_json::from_reader(buffer)
+					.map_err(|_| lightning::ln::msgs::DecodeError::InvalidValue)?;
```
johncantrell97, line 55: There's a bit in the spec about returning jsonrpc::ErrorResponse if we receive an invalid request.  I'm not quite sure how we can do that given this interface.

In theory I'd like to queue the msg here.

Makes me think that we will need to read it as a raw String and handle it up-a-level in the message handler, right?
We can attempt deserialization and then have access to the pending message queue there.

Cargo.toml
```diff
@@ -13,3 +13,6 @@ lightning-invoice = { version = "0.22" }
 lightning-net-tokio = { version = "0.0.114" }
 
 bitcoin = "0.29.2"
+
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
```
TheBlueMatt, line 18: The json bits are optional right? Can we only support lsp0 with a feature flag?

johncantrell97: Well, the entire protocol is defined over jsonrpc abusing a single lightning message as the transport.  So in _theory_ you could receive the raw lightning message payload but you wouldn't be able to do anything with it without json.

src/transport/jsonrpc.rs
```diff
@@ -0,0 +1,161 @@
...
+	/// A number indicating the error type that occurred.
+	pub code: i32,
+	/// A string providing a short description of the error.
+	pub message: String,
+	/// A primitive or structured value that contains additional information about the error.
+	pub data: Option<Value>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+/// A JSONRPC request object.
```
tnull, line 113: nit: Probably want to replace all `JSONRPC` with `JSON-RPC 2.0`.

src/transport/jsonrpc.rs
```diff
@@ -0,0 +1,161 @@
...
+pub struct Request {
+	/// The name of the RPC call.
+	pub method: String,
```
tnull: Could introduce a dedicated `Method` type that holds the prefix, the method name, and a bool indicating whether it's a notification. Then we can also impl conversion traits such as the serde traits and possibly `FromStr`/`Display` on that type.

src/transport/jsonrpc.rs
```diff
@@ -0,0 +1,161 @@
...
+	/// Parameters to the RPC call.
```
tnull, line 117: Could note here that only by-name paramters are allowed.

src/transport/jsonrpc.rs
```diff
@@ -0,0 +1,161 @@
+use std::collections::HashMap;
```
tnull: Missing module docs. Should probably link to https://www.jsonrpc.org/specification

src/transport/jsonrpc.rs
```diff
@@ -0,0 +1,161 @@
...
+/// A JSONRPC request object.
+pub struct Request {
+	/// The name of the RPC call.
+	pub method: String,
+	/// Parameters to the RPC call.
+	pub params: HashMap<String, Value>,
+	/// Identifier for this Request, which should appear in the response.
+	pub id: Value,
+	/// jsonrpc field, MUST be "2.0".
+	pub jsonrpc: String,
```
tnull, line 123: Can we define a `const JSON_RPC_VERSION` for that?

src/transport/mod.rs
```diff
@@ -8,3 +8,7 @@
 // licenses.
 
 //! Types and primitives that implement the LSPS0: Transport Layer specification.
+
+pub mod jsonrpc;
```
tnull, line 12: Ah, also need to mark this mod `pub` in `src/lib.rs`.

src/transport/msgs.rs
```diff
@@ -0,0 +1,65 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct RawLSPSMessage {
+	pub payload: String,
+}
+
+impl_writeable_msg!(RawLSPSMessage, { payload }, {});
+
+impl wire::Type for RawLSPSMessage {
+	fn type_id(&self) -> u16 {
+		37913
```
tnull: Can we make this a crate-wide constant?

src/transport/msgs.rs
```diff
@@ -0,0 +1,65 @@
...
+	}
+}
+
+/// An [`lsps0_message`] message to be sent to or received from a peer.
+///
+/// [`lsps0_message`]: https://github.com/BitcoinAndLightningLayerSpecs/lsp/blob/49544d1a420cf32ef4726c526a29aa46a09202ff/LSPS0/README.md
```
tnull, line 23: Until merged, I'd leave TODOs instead of links so that we can link to the finalized version ASAP (and don't forget to update).

src/transport/msgs.rs
```diff
@@ -0,0 +1,65 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum LSPSMessage {
```
tnull, line 26: Mh, I wonder if there would be value in separating the LSP and the Client side from the start? Given that the roles are pretty clear cut, it might reduce complexity in each of the implementations if we had separate message types and message handlers for each, rather than having one impl that speaks everything. Do you have thoughts on this?

src/transport/message_handler.rs
```diff
@@ -0,0 +1,106 @@
+use crate::transport::{jsonrpc, msgs};
+use bitcoin::secp256k1::PublicKey;
+use lightning::ln::{peer_handler::CustomMessageHandler, wire::CustomMessageReader};
+use std::{io, ops::Deref, sync::Mutex};
+
+pub trait LSPMessageHandler {
+	fn handle_lsps_request(
+		&self, request: jsonrpc::Request, sender_node_id: &PublicKey,
```
tnull: nit: Could name this `counterparty_node_id`, as `sender` can often be ambiguous, especially when dealing with multiple layers of protocols.

src/transport/message_handler.rs
```diff
@@ -0,0 +1,106 @@
...
+			msgs::LSPSMessage::Response(response) => {
+				self.message_handler.handle_lsps_response(response, sender_node_id)
+			}
+			msgs::LSPSMessage::Notification(notification) => {
+				self.message_handler.handle_lsps_notification(notification, sender_node_id)
+			}
+		}
+	}
+
+	fn send_invalid_request(&self, sender_node_id: &PublicKey) {
```
tnull, line 158: I assume this is for testing purposes only? Should we hide it behind the `test` feature, and possibly move it to a module in a `tests` subfolder?

src/transport/message_handler.rs
```diff
@@ -0,0 +1,106 @@
...
+where
+	LMH::Target: LSPMessageHandler,
+{
+	type CustomMessage = msgs::RawLSPSMessage;
+
+	fn read<R: io::Read>(
+		&self, message_type: u16, buffer: &mut R,
+	) -> Result<Option<Self::CustomMessage>, lightning::ln::msgs::DecodeError> {
+		match message_type {
+			37913 => {
```
tnull: Def. need a global constant for this :)

src/transport/message_handler.rs
```diff
@@ -0,0 +1,106 @@
...
+			}
+		}
+	}
+
+	fn send_invalid_request(&self, sender_node_id: &PublicKey) {
+		let response =
+			msgs::LSPSMessage::Response(jsonrpc::Response::Error(jsonrpc::ErrorResponse {
+				jsonrpc: "2.0".to_string(),
+				error: jsonrpc::Error {
+					code: -32600,
```
tnull: Also need a constant for our default error code.

tnull (commented):
Thanks for the initiative! 

After a first pass, I'm not entirely convinced we should expose the Json objects, but rather add helper methods on LSPSMessage to customize them. I think having an interface trait to interact is a good idea though.


> I'm not so sure on this `LSPManager` object. My thought was it handled the layer of going from raw string <==> rpc request/response/notification for the user. It would also expose some api to a higher level and perform whatever bookkeeping is necessary (managing pending request's / ids for them?).

Yes, I'm also not sure yet, we probably still want to keep it flexible. Also, even though it's very early for that, I want to note that we probably want to have some tests checking that our CMH implementation works nicely with [`lightning-custom-message`](https://docs.rs/lightning-custom-message/0.0.114/lightning_custom_message/) to allow users to combine multiple message handlers.

johncantrell97:
> Thanks for the initiative!
> 
> After a first pass, I'm not entirely convinced we should expose the Json objects, but rather add helper methods on LSPSMessage to customize them. I think having an interface trait to interact is a good idea though.
> 

Hm okay.  I agree it's not really necessary for standard jit_channel and channel_request protocols.  I was thinking further down the road the client/lsps might end up with custom protocols or protocol extensions that use jsonrpc and they'd need access to the raw jsonrpc requests to be able to implement it.  I guess it would almost be like the LDK CustomMessageHandler where this library provides handlers for jsonrpc requests that are standard and part of the LSPS and if we receive jsonrpc requests that are non-standard and they provided a CustomJsonRPCRequestHandler or something like that then we could push them to that to handle.

> > I'm not so sure on this `LSPManager` object. My thought was it handled the layer of going from raw string <==> rpc request/response/notification for the user. It would also expose some api to a higher level and perform whatever bookkeeping is necessary (managing pending request's / ids for them?).
> 
> Yes, I'm also not sure yet, we probably still want to keep it flexible. Also, even though it's very early for that, I want to note that we probably want to have some tests checking that our CMH implementation works nicely with [`lightning-custom-message`](https://docs.rs/lightning-custom-message/0.0.114/lightning_custom_message/) to allow users to combine multiple message handlers.

Yeah, it is hard to develop this completely bottom-up so I started another branch to start imagining how I would want to use it as a user of the library (what apis and events I would want to react to and call into) and I think that will help inform how the transport and end-user should connect and what the library needs to handle vs the end-user.

src/transport/jsonrpc.rs
johncantrell97: yeah I think that makes sense to me. I assume at the jit_channel/channel_request layer there'd be a higher level abstraction around the actual request methods and params expected.

src/transport/msgs.rs
johncantrell97: Ah, I hadn't thought of doing that.  It probably makes sense, since most users of this library will be 'clients' who are making requests to the LSP.  I'm a bit biased because our usage of this library is intended to be from the LSP side.

I definitely want to support both use-cases with this library.

I'll keep that in mind and make the split if it seems like it would reduce complexity for the end-users.

src/transport/message_handler.rs
johncantrell97: Hm, actually it was more of a util/helper to not have to construct and queue an error message response in multiple places.  I think maybe it's only used for errors parsing JSON payload right now but I think the spec has this used quite a bit.  

Could also just impl some constructors on LSPSMessage to help DRY it up and have like LSPSMessage::invalid_request_response().

src/transport/message_handler.rs
johncantrell97: Technically, this error code is from JSON-RPC spec

src/transport/jsonrpc.rs
johncantrell97: Started doing this but not sure about including the notification flag? I already had introduced a Notification type that is automatically deserialized when there is no `id` field.

So if you're matching on an LSPSMessage you'd already know if it was a notification or not.  Or are you suggesting removing the Notification type and enum variant.

I kind of like how serde_json can automatically derive Serialize/Deserialize these variants correctly using a combination of `untagged` and `deny_unknown_fields` attributes.

Cargo.toml
TheBlueMatt: Wait what lol. Why? That's disgusting.

johncantrell97: Yeah, I thought it was really strange too. I don't really know what the rationale is or why it doesn't just use multiple protocol messages like lightning does.

tnull: I think the quoted rationale is that it reduces the footprint w.r.t. ports and lightning feature bits, which also means any new sub-protocols wouldn't need individual bLIPs to have them 'registered'. I may add it also enables a clear separation between the transport and higher layers, i.e., the "application layer" (i.e., anything on top of transport) protocols can easily be reused (or proxied) over any of the other discussed/requested transport protocols, such as HTTP. 

I agree that coming from the Lightning protocol side of things this is not necessarily the cleanest design, but have no strong feeling either way for the above mentioned reasons.

TheBlueMatt: I'm fine with a single protocol message, I was more surprised by JSON - none of the advantages of a human readable message on the wire apply (it's all encrypted anyway), so it seemed quite strange that we weren't doing lightning TLVs, Protobuf or some other binary protocol.

johncantrell97:
Ok, just pushed an updated version of this that allows the user to register multiple `LSPMessageHandler`s (one per prefix) to the LSPManager / CustomMessageHandler.

This means each protocol the user wants to support can be registered independently on the custom message handler and messages will be delegated to the appropriate handler based on the prefix.

I extended the trait by adding a `get_and_clear_pending_msg` method so that these protocol message handlers can generate their own custom messages that will get sent to the counterparty.

I also added a `get_protocol_number` method so the LSPS0 handler could implement the `listprotocols` method by iterating the registered handlers and collecting all of the protocol numbers it supports.

src/transport/message_handler.rs
```diff
@@ -0,0 +1,270 @@
...
+	pub fn register_message_handler(&self, prefix: String, message_handler: LMH) {
+		self.message_handlers.lock().unwrap().insert(prefix, message_handler);
+	}
+
+	pub fn list_protocols(
+		&self, counterparty_node_id: PublicKey,
+	) -> Result<(), lightning::ln::msgs::LightningError> {
+		let request = jsonrpc::Request {
+			jsonrpc: jsonrpc::VERSION.to_string(),
+			id: serde_json::json!(1),
```
johncantrell97: TODO: Clients MUST use a high-entropy id string for JSON-RPC 2.0 requests, such as a UUID, or a hex encoding of a random binary blob of at least 80 bits, with randomness acquired from a cryptographically secure source. Clients MUST NOT use just a simple incrementing counter for the id.

johncantrell97: Made handler generic over EntropySource to get random bytes and convert to hex string to use as id.  For now taking first 16 bytes of the 32 byte blob as spec says 80 bits is enough and I think standard UUID is 16 bytes.  Probably still overkill but 🤷

johncantrell97:
I pushed another commit that is an attempt at your idea of hiding jsonrpc types from the protocol level message handlers.  I am finding it extremely awkward to go this route. the commit is not complete but hopefully illustrates some of the pain.

curious to hear if you think this path is really worth going down and if so, any ideas on how to avoid all the pain?

src/transport/message_handler.rs
ZmnSCPxj-jr: Not overkill: none of the implementations actually allow a non-racy way to determine if a custom message is sent before or after a disconnection.

Consider that a disconnection might imply the client crashed and restarted.  If it used an in-memory incrementing counter, then the counter might get reset, meaning the client will duplicate IDs, and there is the possibility that the LSP *then* sends a message (since none of the send-custom-message APIs in any of LDK, LND, and CLN allow non-racy way to ensure sending on the same connection a previous custom message was received on, everybody just keys on node ID) that is for the previous connection session (the one that crashed) which might be confused as the response to a newly-sent request.

Your alternative is a persisted counter (which requires you to actually do `flush`ed disk I/O every time the client makes a request) or a randomly-picked nonce (which just requires a lock on your entropy source and some CPU cycles to update the PRNG).  80 bits is absolute minimum necessary for randomly-picked nonce IMO.  Could get away with 64 bits but I like me nice cheap margins.

tnull: Yeah, given that it has to be unique per connection I'm not sure we need that much entropy. However, given that the most expensive thing about it is probably sourcing the entropy and we do so 32 bytes at a time, it's probably fine to go with 16 bytes.

src/lsps0/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+pub enum LSPS0Response {
+    ListProtocolsSuccess { request_id: String, protocols: Vec<u16> },
+    ListProtocolsError { request_id: String, code: i32, message: String, data: Option<String> },
+    Unknown
+}
+
+impl From<(jsonrpc::Request,jsonrpc::Response)> for LSPS0Response {
```
tnull: Mh, rather then using the `From` trait, couldn't we just implement an `LSPS0Response::from_request(id, name)` constructor or similar?

src/events.rs
```diff
@@ -0,0 +1,10 @@
+use bitcoin::secp256k1::PublicKey;
+
+pub trait EventHandler {
+	fn handle_event(&self, event: Event);
+}
+
+#[derive(Debug, Clone)]
+pub enum Event {
+	ListProtocols { counterparty_node_id: PublicKey, protocols: Vec<u16> },
```
tnull, line 9: We should either introduce an entirely separate type of event (`MsgReceivedEvent`) for messages, or at least make the event name clearly reflect what happened, e.g., `ListProtocolsMsgReceived` or similar.

src/lib.rs
```diff
@@ -11,12 +11,16 @@
...
 #![allow(ellipsis_inclusive_range_patterns)]
+#![allow(clippy::drop_non_drop)]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
 mod channel_request;
+mod events;
 mod jit_channel;
 mod transport;
+mod utils;
+mod lsps0;
```
tnull: Not sure `lsps0` should be separate from `transport`. It **is** the transport layer. If we want to structure everything by `lspsX` prefix, we should have the modules named `lsps0`, `lsps1`, `lsps2` here and remove `transport`,`jit_channel`, and `channel_request`.

src/lsps0/mod.rs
```diff
@@ -0,0 +1,125 @@
+use std::{sync::{Arc, Mutex}, collections::HashMap, ops::Deref};
```
tnull: This style of grouping imports gets confusing really fast, please just use a list and group only based on the most specific import level.

src/lsps0/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+pub enum LSPS0Response {
+    ListProtocolsSuccess { request_id: String, protocols: Vec<u16> },
+    ListProtocolsError { request_id: String, code: i32, message: String, data: Option<String> },
+    Unknown
+}
+
+impl From<(jsonrpc::Request,jsonrpc::Response)> for LSPS0Response {
+    fn from((request, response): (jsonrpc::Request, jsonrpc::Response)) -> Self {
+        match request.method.name.as_str() {
+            "listprotocols" => {
```
tnull: Mh is there a way we could have a `Method` trait in which method name and possibly the request id are associated types? We could then have corresponding implementations for each method that share functionalities?

src/transport/jsonrpc.rs
```diff
@@ -0,0 +1,300 @@
...
+	}
+}
+
+// implement Deserialize for Method using it's FromStr implementation
+impl<'de> Deserialize<'de> for Method {
+	fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+	where
+		D: serde::Deserializer<'de>,
+	{
+		let s = String::deserialize(deserializer)?;
```
tnull, line 97: Can we avoid all the redundant allocation that comes from deserializing everything through `FromStr`?  I'm not sure we need to impl `FromStr` at all for most of these types, but should def. use `Deserialize` to parse directly into corresponding Rust types.

tnull (commented):
Mh, correct me if I'm wrong but from what I can see a good part of the "pain" and needed boilerplate comes exactly from trying to handle JSON objects and dealing with `String`s at multiple places. I still think we should really try to keep the leakage of JSON types to upper layers minimal.

src/lsps0/mod.rs
johncantrell97: hm not exactly because we also need the response. we could do LSPS0Response::from(id, method, response)?

src/lib.rs
johncantrell97: yeah, well the idea was that there's the transport layer and then there's the LSPS0 "protocol".  It currently only has one request (lsps0.listprotocols).

It's not really the transport layer imo.  

Anyway, for now I moved it into a `protocol` module of the transport layer

src/lsps0/mod.rs
johncantrell97: yeah agreed, it's just what my editor does by default.  I fixed these in latest commit.  I wonder if there's a cargo fmt config to do this automatically?

PR #3, opened 2023-05-02
Alternative without intermediate jsonrpc types

johncantrell97:
This is an alternative approach that avoids creating intermediate JSON-RPC objects and deserializes directly into rust types.  It leaks no serde_json::Value's and no need for the json-rpc Request/Response/Notification objects.


src/lib.rs
```diff
@@ -11,12 +11,15 @@
 //! # ldk-lsp-client
 //! Types and primitives to integrate a spec-compliant LSP with an LDK-based node.
 #![deny(missing_docs)]
-#![deny(broken_intra_doc_links)]
-#![deny(private_intra_doc_links)]
+#![deny(rustdoc::broken_intra_doc_links)]
+#![deny(rustdoc::private_intra_doc_links)]
```
tnull, line 15: Btw, I think we should aim for an MSRV of 1.48, just like upstream LDK. Can we enforce this in CI? Noting it here because the change won't compile then I think.

PR #4, opened 2023-05-03
LSPS0 message handling

johncantrell97:
Implements LSPS0 message handling and lays groundwork for other LSPS protocol handling.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,11 +6,16 @@ jobs:
   build:
     strategy:
       matrix:
+        platform: [ ubuntu-latest ]
         toolchain: [ stable, beta ]
         include:
           - toolchain: stable
             check-fmt: true
-    runs-on: ubuntu-latest
+          - toolchain: 1.48.0
+            platform: ubuntu-latest
+            msrv: true
+
+    runs-on: ${{ matrix.platform }}
     steps:
       - name: Checkout source code
         uses: actions/checkout@v2
@@ -20,6 +25,19 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
           override: true
           profile: minimal
+      - name: Pin tokio for MSRV
+        if: matrix.msrv
+        run: cargo update -p tokio --precise "1.14.1" --verbose
+      - name: Pin serde for MSRV
+        if: matrix.msrv
+        run: cargo update -p serde --precise "1.0.156" --verbose
+      - name: Pin log for MSRV
+        if: matrix.msrv
+        run: cargo update -p log --precise "0.4.18" --verbose
+      - name: Cargo check
+        run: cargo check --release
+      - name: Check documentation
+        run: cargo doc --release
       - name: Build on Rust ${{ matrix.toolchain }}
         run: cargo build --verbose --color always
       - name: Check formatting
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,8 +8,11 @@ description = "Types and primitives to integrate a spec-compliant LSP with an LD
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-lightning = { version = "0.0.114", features = ["max_level_trace", "std"] }
-lightning-invoice = { version = "0.22" }
-lightning-net-tokio = { version = "0.0.114" }
+lightning = { git = "https://github.com/lightningdevkit/rust-lightning.git", rev = "498f2331459d8031031ef151a44c90d700aa8c7e", features = ["max_level_trace", "std"] }
+lightning-invoice = { git = "https://github.com/lightningdevkit/rust-lightning.git", rev = "498f2331459d8031031ef151a44c90d700aa8c7e" }
+lightning-net-tokio = { git = "https://github.com/lightningdevkit/rust-lightning.git", rev = "498f2331459d8031031ef151a44c90d700aa8c7e" }
 
-bitcoin = "0.29.2"
+bitcoin = "0.29.0"
+
+serde = { version = "1.0", default-features = false, features = ["derive", "alloc"] }
+serde_json = "1.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,8 +15,12 @@
 #![deny(private_intra_doc_links)]
 #![allow(bare_trait_objects)]
 #![allow(ellipsis_inclusive_range_patterns)]
+#![allow(clippy::drop_non_drop)]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
 mod channel_request;
 mod jit_channel;
 mod transport;
+mod utils;
+
+pub use transport::message_handler::{LiquidityManager, LiquidityProviderConfig};
--- a/src/transport/message_handler.rs
+++ b/src/transport/message_handler.rs
@@ -1,0 +1,165 @@
+use crate::transport::msgs::{LSPSMessage, RawLSPSMessage, LSPS_MESSAGE_TYPE};
+use crate::transport::protocol::LSPS0MessageHandler;
+
+use bitcoin::secp256k1::PublicKey;
+use lightning::ln::features::{InitFeatures, NodeFeatures};
+use lightning::ln::msgs::{ErrorAction, LightningError};
+use lightning::ln::peer_handler::CustomMessageHandler;
+use lightning::ln::wire::CustomMessageReader;
+use lightning::sign::EntropySource;
+use lightning::util::logger::Level;
+use lightning::util::ser::Readable;
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use std::io;
+use std::ops::Deref;
+use std::sync::{Arc, Mutex};
+
+const LSPS_FEATURE_BIT: usize = 729;
+
+/// A trait used to implement a specific LSPS protocol.
+///
+/// The messages the protocol uses need to be able to be mapped
+/// from and into [`LSPSMessage`].
+pub(crate) trait ProtocolMessageHandler {
+	type ProtocolMessage: TryFrom<LSPSMessage> + Into<LSPSMessage>;
+	const PROTOCOL_NUMBER: Option<u16>;
+
+	fn handle_message(
+		&self, message: Self::ProtocolMessage, counterparty_node_id: &PublicKey,
+	) -> Result<(), LightningError>;
+}
+
+/// A configuration for [`LiquidityManager`].
+///
+/// Allows end-user to configure options when using the [`LiquidityManager`]
+/// to provide liquidity services to clients.
+pub struct LiquidityProviderConfig;
+
+/// The main interface into LSP functionality.
+///
+/// Should be used as a [`CustomMessageHandler`] for your
+/// [`lightning::ln::peer_handler::PeerManager`]'s [`lightning::ln::peer_handler::MessageHandler`].
+pub struct LiquidityManager<ES: Deref>
+where
+	ES::Target: EntropySource,
+{
+	pending_messages: Arc<Mutex<Vec<(PublicKey, LSPSMessage)>>>,
+	request_id_to_method_map: Mutex<HashMap<String, String>>,
+	lsps0_message_handler: LSPS0MessageHandler<ES>,
+	provider_config: Option<LiquidityProviderConfig>,
+}
+
+impl<ES: Deref> LiquidityManager<ES>
+where
+	ES::Target: EntropySource,
+{
+	/// Constructor for the LiquidityManager
+	///
+	/// Sets up the required protocol message handlers based on the given [`LiquidityProviderConfig`].
+	pub fn new(entropy_source: ES, provider_config: Option<LiquidityProviderConfig>) -> Self {
+		let pending_messages = Arc::new(Mutex::new(vec![]));
+
+		let lsps0_message_handler =
+			LSPS0MessageHandler::new(entropy_source, vec![], Arc::clone(&pending_messages));
+
+		Self {
+			pending_messages,
+			request_id_to_method_map: Mutex::new(HashMap::new()),
+			lsps0_message_handler,
+			provider_config,
+		}
+	}
+
+	fn handle_lsps_message(
+		&self, msg: LSPSMessage, sender_node_id: &PublicKey,
+	) -> Result<(), lightning::ln::msgs::LightningError> {
+		match msg {
+			LSPSMessage::Invalid => {
+				return Err(LightningError { err: format!("{} did not understand a message we previously sent, maybe they don't support a protocol we are trying to use?", sender_node_id), action: ErrorAction::IgnoreAndLog(Level::Error)});
+			}
+			LSPSMessage::LSPS0(msg) => {
+				self.lsps0_message_handler.handle_message(msg, sender_node_id)?;
+			}
+		}
+		Ok(())
+	}
+
+	fn enqueue_message(&self, node_id: PublicKey, msg: LSPSMessage) {
+		let mut pending_msgs = self.pending_messages.lock().unwrap();
+		pending_msgs.push((node_id, msg));
+	}
+}
+
+impl<ES: Deref> CustomMessageReader for LiquidityManager<ES>
+where
+	ES::Target: EntropySource,
+{
+	type CustomMessage = RawLSPSMessage;
+
+	fn read<R: io::Read>(
+		&self, message_type: u16, buffer: &mut R,
+	) -> Result<Option<Self::CustomMessage>, lightning::ln::msgs::DecodeError> {
+		match message_type {
+			LSPS_MESSAGE_TYPE => Ok(Some(RawLSPSMessage::read(buffer)?)),
+			_ => Ok(None),
+		}
+	}
+}
+
+impl<ES: Deref> CustomMessageHandler for LiquidityManager<ES>
+where
+	ES::Target: EntropySource,
+{
+	fn handle_custom_message(
+		&self, msg: Self::CustomMessage, sender_node_id: &PublicKey,
+	) -> Result<(), lightning::ln::msgs::LightningError> {
+		let mut request_id_to_method_map = self.request_id_to_method_map.lock().unwrap();
+
+		match LSPSMessage::from_str_with_id_map(&msg.payload, &mut request_id_to_method_map) {
+			Ok(msg) => self.handle_lsps_message(msg, sender_node_id),
+			Err(_) => {
+				self.enqueue_message(*sender_node_id, LSPSMessage::Invalid);
+				Ok(())
+			}
+		}
+	}
+
+	fn get_and_clear_pending_msg(&self) -> Vec<(PublicKey, Self::CustomMessage)> {
+		let mut request_id_to_method_map = self.request_id_to_method_map.lock().unwrap();
+		self.pending_messages
+			.lock()
+			.unwrap()
+			.drain(..)
+			.map(|(public_key, lsps_message)| {
+				if let Some((request_id, method_name)) = lsps_message.get_request_id_and_method() {
+					request_id_to_method_map.insert(request_id, method_name);
+				}
+				(
+					public_key,
+					RawLSPSMessage { payload: serde_json::to_string(&lsps_message).unwrap() },
+				)
+			})
+			.collect()
+	}
+
+	fn provided_node_features(&self) -> NodeFeatures {
+		let mut features = NodeFeatures::empty();
+
+		if self.provider_config.is_some() {
+			features.set_optional_custom_bit(LSPS_FEATURE_BIT).unwrap();
+		}
+
+		features
+	}
+
+	fn provided_init_features(&self, _their_node_id: &PublicKey) -> InitFeatures {
+		let mut features = InitFeatures::empty();
+
+		if self.provider_config.is_some() {
+			features.set_optional_custom_bit(LSPS_FEATURE_BIT).unwrap();
+		}
+
+		features
+	}
+}
--- a/src/transport/mod.rs
+++ b/src/transport/mod.rs
@@ -8,3 +8,7 @@
 // licenses.
 
 //! Types and primitives that implement the LSPS0: Transport Layer specification.
+
+pub mod message_handler;
+pub mod msgs;
+pub mod protocol;
--- a/src/transport/msgs.rs
+++ b/src/transport/msgs.rs
@@ -1,0 +1,394 @@
+use lightning::impl_writeable_msg;
+use lightning::ln::wire;
+use serde::de;
+use serde::de::{MapAccess, Visitor};
+use serde::ser::SerializeStruct;
+use serde::{Deserialize, Deserializer, Serialize};
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use std::fmt;
+
+const LSPS_MESSAGE_SERIALIZED_STRUCT_NAME: &str = "LSPSMessage";
+const JSONRPC_FIELD_KEY: &str = "jsonrpc";
+const JSONRPC_FIELD_VALUE: &str = "2.0";
+const JSONRPC_METHOD_FIELD_KEY: &str = "method";
+const JSONRPC_ID_FIELD_KEY: &str = "id";
+const JSONRPC_PARAMS_FIELD_KEY: &str = "params";
+const JSONRPC_RESULT_FIELD_KEY: &str = "result";
+const JSONRPC_ERROR_FIELD_KEY: &str = "error";
+const JSONRPC_INVALID_MESSAGE_ERROR_CODE: i32 = -32700;
+const JSONRPC_INVALID_MESSAGE_ERROR_MESSAGE: &str = "parse error";
+const LSPS0_LISTPROTOCOLS_METHOD_NAME: &str = "lsps0.listprotocols";
+
+pub const LSPS_MESSAGE_TYPE: u16 = 37913;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct RawLSPSMessage {
+	pub payload: String,
+}
+
+impl_writeable_msg!(RawLSPSMessage, { payload }, {});
+
+impl wire::Type for RawLSPSMessage {
+	fn type_id(&self) -> u16 {
+		LSPS_MESSAGE_TYPE
+	}
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct RequestId(pub String);
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct ResponseError {
+	pub code: i32,
+	pub message: String,
+	pub data: Option<String>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize, Default)]
+#[serde(default)]
+pub struct ListProtocolsRequest {}
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct ListProtocolsResponse {
+	pub protocols: Vec<u16>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum LSPS0Request {
+	ListProtocols(ListProtocolsRequest),
+}
+
+impl LSPS0Request {
+	pub fn method(&self) -> &str {
+		match self {
+			LSPS0Request::ListProtocols(_) => "lsps0.listprotocols",
+		}
+	}
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum LSPS0Response {
+	ListProtocols(ListProtocolsResponse),
+	ListProtocolsError(ResponseError),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum LSPS0Message {
+	Request(RequestId, LSPS0Request),
+	Response(RequestId, LSPS0Response),
+}
+
+impl TryFrom<LSPSMessage> for LSPS0Message {
+	type Error = ();
+
+	fn try_from(message: LSPSMessage) -> Result<Self, Self::Error> {
+		match message {
+			LSPSMessage::Invalid => Err(()),
+			LSPSMessage::LSPS0(message) => Ok(message),
+		}
+	}
+}
+
+impl From<LSPS0Message> for LSPSMessage {
+	fn from(message: LSPS0Message) -> Self {
+		LSPSMessage::LSPS0(message)
+	}
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum LSPSMessage {
+	Invalid,
+	LSPS0(LSPS0Message),
+}
+
+impl LSPSMessage {
+	pub fn from_str_with_id_map(
+		json_str: &str, request_id_to_method: &mut HashMap<String, String>,
+	) -> Result<Self, serde_json::Error> {
+		let deserializer = &mut serde_json::Deserializer::from_str(json_str);
+		let visitor = LSPSMessageVisitor { request_id_to_method };
+		deserializer.deserialize_any(visitor)
+	}
+
+	pub fn get_request_id_and_method(&self) -> Option<(String, String)> {
+		match self {
+			LSPSMessage::LSPS0(LSPS0Message::Request(request_id, request)) => {
+				Some((request_id.0.clone(), request.method().to_string()))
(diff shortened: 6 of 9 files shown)


.github/workflows/build.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   build:
     strategy:
       matrix:
-        toolchain: [ stable, beta ]
+        toolchain: [ 1.48, stable, beta ]
```
tnull: There are some changed needed already to make 1.48 work. I think we will at least need to pin `tokio` to 1.14 and do similar with `serde`, as we need to use versions that do not depend on `syn` > 1.x.

See how LDK does it in CI: https://github.com/lightningdevkit/rust-lightning/blob/main/ci/ci-tests.sh

src/transport/message_handler.rs
```diff
@@ -0,0 +1,174 @@
...
+
+	fn handle_message(
+		&self, message: Self::ProtocolMessage, counterparty_node_id: &PublicKey,
+	) -> Result<(), lightning::ln::msgs::LightningError>;
+	fn get_and_clear_pending_protocol_messages(&self) -> Vec<(PublicKey, Self::ProtocolMessage)>;
+	fn get_and_clear_pending_protocol_events(&self) -> Vec<Event>;
+	fn get_protocol_number(&self) -> Option<u16>;
+}
+
+pub trait MessageHandler {
```
tnull: Could you add some docs to clarify what difference between protocol and non-protocol messages / message handlers are? Maybe prefix this `MessageHandler` with `Transport` or similar?

src/transport/message_handler.rs
```diff
@@ -0,0 +1,174 @@
...
+
+pub trait ProtocolMessageHandler {
+	type ProtocolMessage: TryFrom<LSPSMessage> + Into<LSPSMessage>;
+
+	fn handle_message(
+		&self, message: Self::ProtocolMessage, counterparty_node_id: &PublicKey,
+	) -> Result<(), lightning::ln::msgs::LightningError>;
+	fn get_and_clear_pending_protocol_messages(&self) -> Vec<(PublicKey, Self::ProtocolMessage)>;
+	fn get_and_clear_pending_protocol_events(&self) -> Vec<Event>;
+	fn get_protocol_number(&self) -> Option<u16>;
```
tnull: I believe this could be a default impl. and the number an associated type?

src/transport/message_handler.rs
```diff
@@ -0,0 +1,174 @@
...
+			}
+		}
+	}
+
+	fn handle_lsps_message(
+		&self, msg: LSPSMessage, sender_node_id: &PublicKey,
+	) -> Result<(), lightning::ln::msgs::LightningError> {
+		if let Some(prefix) = msg.prefix() {
+			let message_handlers = self.message_handlers.lock().unwrap();
+			// TODO: not sure what we are supposed to do when we receive a message we don't have a handler for
```
tnull: I think the default should be to log and ignore unknown messages, if they don't violate LSPS0 requirements.

src/transport/msgs.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+
+impl Serialize for LSPSMessage {
+	fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+	where
+		S: serde::Serializer,
+	{
+		let mut jsonrpc_object = serializer.serialize_struct("LSPSMessage", 3)?;
+
+		jsonrpc_object.serialize_field("jsonrpc", "2.0")?;
```
tnull: Can we make all of these static strings constants to allow type-checking them?

src/transport/msgs.rs
```diff
@@ -0,0 +1,384 @@
...
+						jsonrpc_object.serialize_field("result", result)?;
+					}
+					LSPS0Response::ListProtocolsError(error) => {
+						jsonrpc_object.serialize_field("error", error)?;
+					}
+				}
+			}
+			LSPSMessage::Invalid => {
+				let error =
+					ResponseError { code: -32700, message: "parse error".to_string(), data: None };
```
tnull: Similar with this error code: should be a constant.

src/transport/msgs.rs
```diff
@@ -0,0 +1,384 @@
...
+				_ => {
+					let _: serde_json::Value = map.next_value()?;
+				}
+			}
+		}
+
+		if let Some(method) = method {
+			if let Some(id) = id {
+				match method {
+					"lsps0.listprotocols" => {
```
tnull: Could introduce constants for all of these these method names, too.

src/transport/msgs.rs
```diff
@@ -0,0 +1,384 @@
...
+						)))
+					}
+					_ => Err(de::Error::custom(format!(
+						"Received request with unknown method: {method}"
+					))),
+				}
+			} else {
+				Err(de::Error::custom(format!("Received unknown notification: {method}")))
+			}
+		} else if let Some(id) = id {
```
tnull: Mh, can we try to reduce indentation levels here to improve readability? I think there are multiple way to do this, for example matching on `(id, method)`. Also, it would help to add some comments  and/or nicely named  boolean variables to convey what our parsing state is currently.

src/transport/msgs.rs
```diff
@@ -0,0 +1,384 @@
...
+			} else {
+				Err(de::Error::custom(format!("Received response for unknown request id: {id}")))
+			}
+		} else {
+			Err(de::Error::custom("Received invalid JSON-RPC object: one of method or id required"))
+		}
+	}
+}
+
+#[cfg(test)]
```
tnull, line 271: I think it may be nice to add property-based testing based on the `proptest` crate here in a follow-up. This would improve test-coverage and reduce the chance we're parsing bogus random inputs.

johncantrell97:
Ok, addressed all of your comments and updated CI to run against 1.48.0 successfully.

src/lib.rs
```diff
@@ -15,8 +15,11 @@
 #![deny(private_intra_doc_links)]
 #![allow(bare_trait_objects)]
 #![allow(ellipsis_inclusive_range_patterns)]
+#![allow(clippy::drop_non_drop)]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
 mod channel_request;
+mod events;
 mod jit_channel;
 mod transport;
```
tnull, line 23: We should expose this module and make sure that `cargo doc` is successful (currently a few exposed methods are missing documentation I believe).

.github/workflows/build.yml
```diff
@@ -20,6 +25,12 @@ jobs:
...
           profile: minimal
+      - name: Pin tokio for MSRV
+        if: matrix.msrv
+        run: cargo update -p tokio --precise "1.14.0" --verbose
+      - name: Pin serde for MSRV
+        if: matrix.msrv
+        run: cargo update -p serde --precise "1.0.156" --verbose
       - name: Build on Rust ${{ matrix.toolchain }}
         run: cargo build --verbose --color always
       - name: Check formatting
```
tnull, line 43: Can we add a step with `cargo check --release` and `cargo doc --release`, to make sure no exposed methods are undocumented? We possibly should even add another step for `cargo doc --document-private-items`, to enforce the same for non-exposed methods, but I'll leave that up to you.

src/transport/mod.rs
```diff
@@ -8,3 +8,7 @@
 // licenses.
 
 //! Types and primitives that implement the LSPS0: Transport Layer specification.
+
+pub mod message_handler;
```
tnull, line 12: These modules are missing docs.

src/transport/message_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+use lightning::ln::wire::CustomMessageReader;
+use lightning::log_info;
+use lightning::util::logger::Logger;
+use std::collections::HashMap;
+use std::convert::{TryFrom, TryInto};
+use std::io;
+use std::ops::Deref;
+use std::sync::{Arc, Mutex};
+
+/// A trait used to implement a specific LSPS protocol
```
tnull: Mh, I think this needs much better docs: how does this relate to `TransportMessageHandler` exactly? Since `ProtocolMessageHandler` is a general trait that *each* protocol needs to implement, should this rather live in the general namespace, for example directly in`lib.rs` for the time being?

nit: Let's keep the first descriptive line separate, which improves display in module-level docs.
```suggestion
/// A trait used to implement a specific LSPS protocol.
///
```

src/transport/message_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+/// The messages the protocol uses need to be able to be mapped
```
tnull, line 22: This line seems to belong somewhere else?

src/transport/message_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+/// from and into LSPSMessages.
```
tnull: nit: let's tick and link all non-local objects here and elsewhere (i.e., everything that you're currently not describing should be linked)
```suggestion
/// from and into [`LSPSMessages`].
```

src/transport/message_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+		&self, message: Self::ProtocolMessage, counterparty_node_id: &PublicKey,
+	) -> Result<(), lightning::ln::msgs::LightningError>;
+	fn get_and_clear_pending_protocol_messages(&self) -> Vec<(PublicKey, Self::ProtocolMessage)>;
+	fn get_and_clear_pending_protocol_events(&self) -> Vec<Event>;
+	fn get_protocol_number(&self) -> Option<u16> {
+		Self::PROTOCOL_NUMBER
+	}
+}
+
+/// A trait used to implement the mapping from a LSPS transport layer mesage
```
tnull:
```suggestion
/// A trait used to implement the mapping from a LSPS transport layer message
```

src/transport/message_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+/// to a specific protocol message. This enables the ProtocolMessageHandler's
+/// to not need to know about LSPSMessage and only have to deal with the specific
+/// messages related to the protocol that is being implemented.
+pub trait TransportMessageHandler {
```
tnull: I'm a bit dubious why we need this second trait exactly? Since it's not used in the LSPS2 PR, is it just there for the blanket impl? Couldn't the blanket impl be added to `ProtocolMessageHandler`? 

If we need it somehow, should we be adding the type constraint here?
```suggestion
pub trait TransportMessageHandler: ProtocolMessageHandler {
```

src/transport/message_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+	fn get_and_clear_pending_events(&self) -> Vec<Event> {
+		self.get_and_clear_pending_protocol_events()
+	}
+
+	fn get_protocol_number(&self) -> Option<u16> {
+		self.get_protocol_number()
+	}
+}
+
+pub struct LSPManager<L: Deref>
```
tnull: This also needs docs. Also the name doesn't seem to really fit, as, IIUC, this is really just the message handler for LSPS0 messages that does serialization and delegates upwards to the respective `ProtocolMessageHandlers`?

src/transport/message_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+	L::Target: Logger,
+{
+	type CustomMessage = RawLSPSMessage;
+
+	fn read<R: io::Read>(
+		&self, message_type: u16, buffer: &mut R,
+	) -> Result<Option<Self::CustomMessage>, lightning::ln::msgs::DecodeError> {
+		match message_type {
+			LSPS_MESSAGE_TYPE => {
+				let mut payload = String::new();
```
tnull: If we're using `String`, lets initialize it at least via `String::with_capacity` to avoid having to resize upon read.

src/transport/message_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+	) -> Result<(), lightning::ln::msgs::LightningError> {
+		let mut request_id_to_method_map = self.request_id_to_method_map.lock().unwrap();
+
+		match LSPSMessage::from_str_with_id_map(&msg.payload, &mut request_id_to_method_map) {
+			Ok(msg) => self.handle_lsps_message(msg, sender_node_id),
+			Err(_) => {
+				self.enqueue_message(
+					*sender_node_id,
+					RawLSPSMessage {
+						payload: serde_json::to_string(&LSPSMessage::Invalid).unwrap(),
```
tnull: Can this `unwrap` get hit if someone sends us bogus data? Let's we avoid any `unwrap` and panicking in general.

src/transport/message_handler.rs
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}
+
+	pub fn get_message_handlers(
+		&self,
+	) -> Arc<Mutex<HashMap<Prefix, Arc<dyn TransportMessageHandler>>>> {
+		self.message_handlers.clone()
+	}
+
+	pub fn register_message_handler(
```
tnull: Is the plan to expose this registration interface to users? Otherwise, if we're handling a clearly defined set of supported LSPSs anyways, shouldn't we just add them manually, which may also allow us to get rid of the boilerplate around iterating over `Vec`s of message handlers. As it would be only a really small number, could we then just go ahead and check each supported message handler manually, which may also be much more readable?

src/transport/msgs.rs
```diff
@@ -0,0 +1,406 @@
...
+const LSPS0_LISTPROTOCOLS_METHOD_NAME: &str = "lsps0.listprotocols";
+
+pub const LSPS_MESSAGE_TYPE: u16 = 37913;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct RawLSPSMessage {
+	pub payload: String,
+}
+
+impl_writeable_msg!(RawLSPSMessage, { payload }, {});
```
tnull, line 30: Why not make this TLV?

src/transport/msgs.rs
```diff
@@ -0,0 +1,406 @@
...
+					let _: serde_json::Value = map.next_value()?;
+				}
+			}
+		}
+
+		match (id, method) {
+			(Some(id), Some(method)) => match method {
+				LSPS0_LISTPROTOCOLS_METHOD_NAME => {
+					let list_protocols_request =
+						serde_json::from_value(params.unwrap_or(json!({})))
```
tnull: This looks infallible to me due to the `unwrap_or`? But then `list_protocols_request` would be empty? Then again, `params` seems to be required `Some`?

src/transport/msgs.rs
```diff
@@ -0,0 +1,406 @@
...
+					)))
+				}
+				_ => Err(de::Error::custom(format!(
+					"Received request with unknown method: {}",
+					method
+				))),
+			},
+			(Some(id), None) => {
+				if let Some(method) = self.request_id_to_method.get(&id) {
+					match method.as_str() {
```
tnull: Mh, can we avoid the allocation just for comparison's purposes? Also, might be able to make this a bit readable by combining the `if` and `match`.

src/transport/protocol.rs
```diff
@@ -0,0 +1,271 @@
...
+		Self {
+			logger,
+			message_handlers,
+			pending_messages: Mutex::new(vec![]),
+			entropy_source,
+			pending_events: Mutex::new(vec![]),
+		}
+	}
+
+	pub fn generate_request_id(&self) -> RequestId {
```
tnull: Why is this pub? Are we expecting other message handlers to reuse this?

src/transport/protocol.rs
```diff
@@ -0,0 +1,271 @@
...
+				self.enqueue_event(Event::ListProtocols {
+					counterparty_node_id: *counterparty_node_id,
+					protocols,
+				});
+				Ok(())
+			}
+			LSPS0Response::ListProtocolsError(ResponseError { code, message, data, .. }) => {
+				log_error!(
+					self.logger,
+					"Received error({}) response to listprotocols with message: {} data: {:?}",
```
tnull: I'm a bit dubious if we should log in the MessageHandlers themselves, or only propagate errors up and do systematic logging in LSPManager. Thoughts on this?

tnull (commented):
Sorry for the delay here. Generally it's great that we're separating layers now, but I'm still a bit confused about the current design.

In particular I'm not entirely sure if there is no way around splitting `TransportMessageHandler` and `ProtocolMessageHandler`? Also, all of this needs docs, which would make it a lot clearer and would also help think about it further.

src/transport/protocol.rs
```diff
@@ -0,0 +1,271 @@
...
+};
+
+pub struct LSPS0MessageHandler<L: Deref, ES: Deref>
+where
+	L::Target: Logger,
+	ES::Target: EntropySource,
+{
+	logger: L,
+	message_handlers: Arc<Mutex<HashMap<Prefix, Arc<dyn TransportMessageHandler>>>>,
+	pending_messages: Mutex<Vec<(PublicKey, LSPS0Message)>>,
```
tnull: Can't we pass around an `Arc<Mutex<..>>` to the `LSPManager`'s `pending_messages` and have each message handler push directly to that queue?

Cargo.toml
tnull, line 1: I think we should rebase this on the latest revision of upstream main (note: a specific revision, not `main` branch itself) to include https://github.com/lightningdevkit/rust-lightning/pull/2204, i.e., the ability for us to advertise the custom feature bits right away, without the need to add additional `RoutingMessageHandler` boilerplate.

src/transport/msgs.rs
johncantrell97: Yeah, I mean this particular method is odd because it has no parameters.  So we could actually just ignore the `params` key entirely and just construct the message as soon as we match the method name and ensure it has an id.

I think I was mostly just doing it to think through how you'd do it if your method _did_ require params but maybe I'll leave that for the LSPS2 PR

src/transport/msgs.rs
johncantrell97: where is the allocation happening? 

do you mean the String we put in the `request_id_to_method` map? I think it requires an allocation, right? If you wanted to store a &str in the map the actual string would need to live somewhere, no?

.github/workflows/build.yml
```diff
@@ -20,6 +25,14 @@ jobs:
...
           override: true
           profile: minimal
+      - name: Pin tokio for MSRV
+        if: matrix.msrv
+        run: cargo update -p tokio --precise "1.14.0" --verbose
+      - name: Pin serde for MSRV
+        if: matrix.msrv
+        run: cargo update -p serde --precise "1.0.156" --verbose
+      - name: Check documentation
+        run: cargo check --release && cargo doc --release
```
tnull: Checking and building docs are two separate steps and hence should be two separate CI steps too.

Cargo.toml
```diff
@@ -8,8 +8,11 @@ description = "Types and primitives to integrate a spec-compliant LSP with an LD
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-lightning = { version = "0.0.114", features = ["max_level_trace", "std"] }
-lightning-invoice = { version = "0.22" }
-lightning-net-tokio = { version = "0.0.114" }
+lightning = { git = "https://github.com/lightningdevkit/rust-lightning.git", rev = "498f2331459d8031031ef151a44c90d700aa8c7e", features = ["max_level_trace"] }
```
tnull: Let's stick with `std` for now if we're not going no_std anyways.

src/events.rs
```diff
@@ -0,0 +1,28 @@
+// This file is Copyright its original authors, visible in version control
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! Events are returned from various bits in the library which indicate some action must be taken
```
tnull: Not sure what "various bits" is referring to here. Can we make this more concrete? Also, as I read it it means that the user **must** handle all events and take action? Or are we assuming some events to be mostly informational?

src/events.rs
```diff
@@ -0,0 +1,28 @@
...
+//! by the client.
+//!
+//! Because we don't have a built-in runtime, it's up to the client to call events at a time in the
```
tnull: I'm not sure what "call events" means here?

src/events.rs
```diff
@@ -0,0 +1,28 @@
...
+//! future.
+use bitcoin::secp256k1::PublicKey;
+
+/// An Event which you should probably take some action in response to.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Event {
+	/// A response to a ListProtocols request that informs the caller
```
tnull: How are we assuming the user should act in response to receiving an `ListProtocols` event? Do we event want/need to expose this?

src/transport/message_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+
+	fn handle_message(
+		&self, message: Self::ProtocolMessage, counterparty_node_id: &PublicKey,
+	) -> Result<(), lightning::ln::msgs::LightningError>;
+}
+
+/// Configuration for the LSPManager
+/// Allows end-user to configure the behavior of the client
+pub struct LSPConfig {
+	/// whether or not you intend to provide services
```
tnull:
```suggestion
	/// Indicates whether or not you intend to provide services.
```

src/transport/message_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+pub(crate) trait ProtocolMessageHandler {
+	type ProtocolMessage: TryFrom<LSPSMessage> + Into<LSPSMessage>;
+	const PROTOCOL_NUMBER: Option<u16>;
+
+	fn handle_message(
+		&self, message: Self::ProtocolMessage, counterparty_node_id: &PublicKey,
+	) -> Result<(), lightning::ln::msgs::LightningError>;
+}
+
+/// Configuration for the LSPManager
```
tnull:
```suggestion
/// A configuration for [`LSPManager`]
///
```

src/transport/message_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+/// Allows end-user to configure the behavior of the client
```
tnull: I think `client` is confusing terminology here (as LSPS discern 'client' and 'LSP'). Also not sure what additional information this sentence conveys.

src/transport/message_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+pub struct LSPConfig {
+	/// whether or not you intend to provide services
+	is_lsp: bool,
+}
+
+/// The main interface into LSP functionality
```
tnull: Can we make these docs more verbose? What is the user supposed to do with this object? Should he register it somehow somewhere? Does he need to call methods regularly?

src/transport/message_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+impl<L: Deref, ES: Deref> LSPManager<L, ES>
+where
+	L::Target: Logger,
+	ES::Target: EntropySource,
+{
+	/// Constructor for the LSPManager
+	///
+	/// Sets up all required protocol message handlers based on configuration passed in
```
tnull:
```suggestion
	/// Sets up the required protocol message handlers based on the given [`LSPConfig`].
```

src/transport/message_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+			pending_events,
+			request_id_to_method_map: Mutex::new(HashMap::new()),
+			lsps0_message_handler,
+			config,
+		}
+	}
+
+	/// List the LSP protocols a node supports
+	///
+	/// Will receive the response via a ListProtocols event
```
tnull: As said above, I'm not entirely sure we should expose this, and not just call and handle it internally upon first connection? If we want the user to be able to query the supported protocols per-counterparty we should probably store the response in an `Option` field as part of a `LSPSPeer` metadata struct or similar and provide a getter rather than exposing it via an event type.

```suggestion
	/// Will receive the response via a [`ListProtocols`] event
```

src/transport/message_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+	pub fn list_protocols(&self, counterparty_node_id: PublicKey) {
+		self.lsps0_message_handler.list_protocols(counterparty_node_id)
+	}
+
+	/// Needs to be polled regularly to surface events generated by the various protocols
```
tnull: nit: 
```suggestion
	/// Needs to be polled regularly to surface events generated by message handlers.
```

src/transport/message_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+	///
+	/// Sets up all required protocol message handlers based on configuration passed in
+	pub fn new(logger: L, entropy_source: ES, config: LSPConfig) -> Self {
+		let pending_messages = Arc::new(Mutex::new(vec![]));
+		let pending_events = Arc::new(Mutex::new(vec![]));
+
+		let lsps0_message_handler = LSPS0MessageHandler::new(
+			entropy_source,
+			vec![],
+			pending_messages.clone(),
```
tnull: Just a stylistic nit, but I much prefer to use `Arc::clone(&pending_messages)` over `pending_messages.clone()` as it immediately tells that no actual deep-copy is happening here.

src/transport/message_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+	pub fn get_and_clear_pending_events(&self) -> Vec<Event> {
+		self.pending_events.lock().unwrap().drain(..).collect()
+	}
+
+	fn handle_lsps_message(
+		&self, msg: LSPSMessage, sender_node_id: &PublicKey,
+	) -> Result<(), lightning::ln::msgs::LightningError> {
+		match msg {
+			LSPSMessage::Invalid => {
+				log_info!(self.logger, "Received invalid message from {:?}", sender_node_id);
```
tnull: I'm confused: wouldn't the receipt of `LSPSMessage::Invalid` mean that we *sent* an invalid message and the peer responded with `LSPSMessage::Invalid`?

Even if we're not returning an error, should we still log it as an error? Plus, `PublicKey` implements `Display`:
```suggestion
				log_error!(self.logger, "Received invalid message from {}", sender_node_id);
```

src/transport/protocol.rs
```diff
@@ -0,0 +1,242 @@
...
+{
+	pub fn new(
+		entropy_source: ES, protocols: Vec<u16>,
+		pending_messages: Arc<Mutex<Vec<(PublicKey, LSPSMessage)>>>,
+		pending_events: Arc<Mutex<Vec<Event>>>,
+	) -> Self {
+		Self { entropy_source, protocols, pending_messages, pending_events }
+	}
+
+	fn generate_request_id(&self) -> RequestId {
```
tnull: As discussed offline, this should probably be a crate-level helper function.

src/transport/protocol.rs
```diff
@@ -0,0 +1,242 @@
...
+		match response {
+			LSPS0Response::ListProtocols(ListProtocolsResponse { protocols }) => {
+				self.enqueue_event(Event::ListProtocols {
+					counterparty_node_id: *counterparty_node_id,
+					protocols,
+				});
+				Ok(())
+			}
+			LSPS0Response::ListProtocolsError(ResponseError { code, message, data, .. }) => {
+				Err(LightningError {
```
tnull, line 71: Mh, so for some errors we're now ignoring and  logging ourselves, and for some we return `ErrorAction`. I think both are valid, but we should make sure we're using them purposefully (not saying we aren't but we should check).

src/transport/msgs.rs
tnull: Mh, could have sworn there was a `to_string` somewhere in there, but nevermind then for now.

.github/workflows/build.yml
johncantrell97: ok, don't we already essentially cover the `check` when we build after? Or you want to explicitly check --release but not build --release?

Cargo.toml
johncantrell97: ok, I was going to let you help me support no_std after this is merged.  so I'll put it back for now

src/events.rs
johncantrell97: Heh I just copied that comment from the events module of LDK itself.

I guess there is this recurring "problem" where this library is for both the "client" and the "lsp".  So each side will only need to care about the events relevant to it.  

This is also true with the Config and the LSPManager API exposed methods.  

At the moment all relevant events must be handled in some way.

src/events.rs
johncantrell97: yeah will reword this with correct information

src/events.rs
johncantrell97: yeah I guess currently there's really no reason to use it.  I guess the idea is in a world where your wallet lets you add arbitrary LSPs it could use this to surface which ones support JIT Channels or whatever feature you are interested in using.

you could also just try to issue the protocol request and you'll get an error back without ever using list protocols.

should I just remove it for now?

src/transport/message_handler.rs
johncantrell97: Ah, I was never really sure why people used Arc::clone.  That makes sense

src/transport/message_handler.rs
johncantrell97: yeah that's true, message is confusing. fixing it

src/transport/protocol.rs
johncantrell97: Well, you had a previous comment about not logging in the protocol message handlers and bubbling them up.  Now in lsps0 and lsps2 message handlers I removed the dependency on a logger and return ErrorAction's that IgnoreAndLog.

I see there's still the one log_error from above in the root message handler. I'll switch that to the ErrorAction as well.  I assume LDK will log these for me?

With this change we can remove the Logger dependency entirely

src/events.rs
tnull: > yeah I guess currently there's really no reason to use it. I guess the idea is in a world where your wallet lets you add arbitrary LSPs it could use this to surface which ones support JIT Channels or whatever feature you are interested in using.
> 
> you could also just try to issue the protocol request and you'll get an error back without ever using list protocols.

Yeah, I think in this ideal world all LSPs support all LSPSs and we just handle all these protocol specific stuff internally.

> should I just remove it for now?

Yes, I think so. As said before if we want to expose it, I'd do it in another way.

src/transport/protocol.rs
tnull: > Well, you had a previous comment about not logging in the protocol message handlers and bubbling them up. Now in lsps0 and lsps2 message handlers I removed the dependency on a logger and return ErrorAction's that IgnoreAndLog.
> 
> I see there's still the one log_error from above in the root message handler. I'll switch that to the ErrorAction as well. I assume LDK will log these for me?

Yes, it should. But better double-check, not sure if CustomMsgHandler is a special case.

> 
> With this change we can remove the Logger dependency entirely

Mh, why not for the time being.

.github/workflows/build.yml
tnull: Yes, I think we generally want to build in `debug` to have CI not take ages, but should still `check` that it will build in `release`.

src/transport/message_handler.rs
tnull: Want to add that this doubly confusing as the naming `LSPConfig` conveys this is the config .. of an LSP. However, we're then talking about the 'client', which can in our case be both LSP- and client-side? I think we should really clean up terminology here. We should even consider to drop the `LSP` config from the objects as everything here is `LSPsomething` and go for more descriptive object names.

johncantrell97:
I made the minor nit doc changes and linked where made sense.  I added a little bit more information to LSPManager around expected usage.  The problem is that in this version there's really no real use without any of the protocols added.  There's no events or even public api to use.

The only outstanding "issue" imo is naming.  The only exposed objects are the LSPConfig and LSPManager.  "LSP" typically refers to the server, the one doing the providing.  I think you said the spec uses "lsp" and "client".  

The challenge is currently there is a singular config/'manager' for both use cases.  I don't think the LSP prefixed words are the worst but am open to suggestions if you have some.

I could see maybe moving to using a word like Liquidity instead of LSP but I think an "LSP" encompasses more than just liquidity.

tnull:
I think we need to incorporate this fix also to unbreak CI: https://github.com/lightningdevkit/rust-lightning/pull/2353

tnull (commented):
LGTM for now, let us land this and address anything else in follow-ups.

Please squash!

PR #7, opened 2023-06-15
Scaffolding for event handling in future protocols

johncantrell97:
Adds the concept of events to the library.  Future protocols should:

- add Event enum variants
- give protocol message handler a reference to the `pending_events` EventQueue.
- push any relevant events as they are generated inside their protocol

End user is currently expected to poll for events using either:

- `get_and_clear_pending_events` in order to retrieve all currently queued events without blocking (returns empty vec if nothing present)

or

- `get_next_event` blocks until next event is ready and returns it

Perhaps we want an async version that returns a Future as well?

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -1,0 +1,58 @@
+// This file is Copyright its original authors, visible in version control
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! Events are surfaced by the library to indicate some action must be taken
+//! by the end-user.
+//!
+//! Because we don't have a built-in runtime, it's up to the end-user to poll
+//! [`crate::LiquidityManager::get_and_clear_pending_events()`] to receive events.
+
+use std::collections::VecDeque;
+use std::sync::{Condvar, Mutex};
+
+#[derive(Default)]
+pub(crate) struct EventQueue {
+	queue: Mutex<VecDeque<Event>>,
+	condvar: Condvar,
+}
+
+impl EventQueue {
+	pub fn enqueue(&self, event: Event) {
+		{
+			let mut queue = self.queue.lock().unwrap();
+			queue.push_back(event);
+		}
+
+		self.condvar.notify_one();
+	}
+
+	pub fn wait_next_event(&self) -> Event {
+		let mut queue =
+			self.condvar.wait_while(self.queue.lock().unwrap(), |queue| queue.is_empty()).unwrap();
+
+		let event = queue.pop_front().expect("non empty queue");
+		let should_notify = !queue.is_empty();
+
+		drop(queue);
+
+		if should_notify {
+			self.condvar.notify_one();
+		}
+
+		event
+	}
+
+	pub fn get_and_clear_pending_events(&self) -> Vec<Event> {
+		self.queue.lock().unwrap().drain(..).collect()
+	}
+}
+
+/// Event which you should probably take some action in response to.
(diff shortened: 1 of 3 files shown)


src/events.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+impl EventQueue {
+	pub fn push(&self, event: Event) {
+		let mut queue = self.queue.lock().unwrap();
+		queue.push_back(event);
+		self.condvar.notify_one();
+	}
+
+	pub fn pop(&self) -> Event {
```
tnull: May I suggest to name these methods something along the line of `add_event`/`enqueue` and `wait_next_event`, the latter to indicate this will yield and block the current thread?

src/events.rs
```diff
@@ -0,0 +1,53 @@
...
+		let mut queue = self.queue.lock().unwrap();
+
+		let event = loop {
```
tnull: I think instead of implementing our own loop we should just use `Condvar::wait_while`.

src/transport/message_handler.rs
```diff
@@ -65,12 +67,27 @@ where
...
 			request_id_to_method_map: Mutex::new(HashMap::new()),
 			lsps0_message_handler,
 			provider_config,
 		}
 	}
 
+	/// Blocks until next event is ready and returns it
+	///
+	/// Typically you would spawn a thread or task that calls this in a loop
+	pub fn get_next_event(&self) -> Event {
```
tnull: Same here, would much prefer `wait_next_event` to clearly indicate this will block the thread.

src/events.rs
```diff
@@ -0,0 +1,53 @@
...
+		drop(queue);
+		event
+	}
+
+	pub fn get_and_clear_pending_events(&self) -> Vec<Event> {
+		self.queue.lock().unwrap().drain(..).collect()
+	}
+}
+
+/// An Event which you should probably take some action in response to.
```
tnull: nit:
```suggestion
/// An event which you should probably take some action in response to.
```

(alternatively `Event`)

src/events.rs
```diff
@@ -0,0 +1,53 @@
...
+pub(crate) struct EventQueue {
+	queue: Mutex<VecDeque<Event>>,
+	condvar: Condvar,
+}
+
+impl EventQueue {
+	pub fn push(&self, event: Event) {
+		let mut queue = self.queue.lock().unwrap();
+		queue.push_back(event);
+		self.condvar.notify_one();
```
tnull, line 32: I think we should drop the lock before notifying here, otherwise the blocked thread might be woken, but can't acquire the lock right away, i.e., internally possibly doing some busy waiting.

src/events.rs
```diff
@@ -0,0 +1,53 @@
...
+		let mut queue = self.queue.lock().unwrap();
+
+		let event = loop {
+			if let Some(event) = queue.pop_front() {
+				break event;
+			} else {
+				queue = self.condvar.wait(queue).unwrap();
+			}
+		};
+		drop(queue);
```
tnull, line 42: Mh, suppose we push two events A and B, but two calls to `pop` in separate threads would be made at roughly the same time, a sequence something like this: 
```
push(A) | push(B) 
                    T1.pop() 
                    T2.pop()
```
We would expect the two threads to each get an event, but I think this isn't necessarily true: when one of `T1.pop()`  and `T2.pop()` acquires the `Mutex`, the other one would yield and block on `wait`. However, at this point it would never been woken/notified again as no `notify` call is issued again.

I think the notification should have the semantics of "look, something changed and the queue is non-empty", i.e., we here want to a) get a bool if the queue is empty after popping the event b) drop the queue lock c) notify if the event to-be-returned wasn't the last one in the queue.

tnull (commented):
I also wonder if we need to persist this event queue? What happens when the node is shutdown and the users hasn't handled all events. Is the state of all protocols really so stateless that everything will start over? If not and we don't persist it, we need to make sure in each LSPS that every necessary event is re-generated on startup.

tnull (approved):
LGTM.

PR #8, opened 2023-06-20
LSPS1- Channel request

Srg213:


src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,892 @@
...
+#[derive(PartialEq)]
+enum ChannelState {
+	InfoRequested,
+	OrderRequested,
+	PendingSelection,
+	PendingPayment,
+	Ready,
+}
+
+struct CRchannel {
```
tnull: nit: Should name this `CRChannel` probably.
```suggestion
struct CRChannel {
```

src/channel_request/utils.rs
```diff
@@ -0,0 +1,8 @@
+
+// Need to add more conditions
+
+pub fn check_if_valid(value: u128, upper: u128, lower: u128) -> () {
```
tnull: This should probably return an error or a bool and needs a much more descriptive name. Also, I'm not sure if this really warrants a util function and wouldn't be more readable if just inlined.

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,892 @@
...
+	channel_id: u128,
```
tnull: Channel IDs in LDK are usually `[u8; 32]`, we should follow this convention here.

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,892 @@
...
+		self.request_to_cid.insert(request_id, channel_id);
+	}
+
+	pub fn get_channel_in_state_for_request(
+		&mut self, request_id: &RequestId, state: ChannelState,
+	) -> Option<&mut CRchannel> {
+		let channel_id = self.request_to_cid.remove(request_id)?;
+
+		if let Some(channel) = self.channels_by_id.get_mut(&channel_id) {
+			if channel.state == state {
```
tnull: Mh, IIUC this means we would just remove the channel and fail silently if the channel is in a wrong state? We should probably return an error and at least log this, even though it should never happen.

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,892 @@
...
+		}
+	}
+
+	pub fn set_peer_manager(
+		&self, peer_manager: Arc<PeerManager<Descriptor, CM, RM, OM, L, CMH, NS>>,
+	) {
+		*self.peer_manager.lock().unwrap() = Some(peer_manager);
+	}
+
+	fn connect_to_counterparty(&self, counterparty_node_id: PublicKey) {
```
tnull: I'm confused: what is the purpose of this method and why are we generating a random channel id?

tnull (commented):
This needs a rebase after #7 has been merged. Also, compilation currently fails in multiple places.

src/events.rs
tnull, line 1: Please avoid including the merge commit and instead rebase your branch on top of the most-recent `main`.

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,834 @@
+use std::collections::HashMap;
+use std::marker::Sync;
+use std::ops::Deref;
+use std::sync::{Arc, Mutex, RwLock};
+//use std::time::{self, Instant};
```
tnull: nit: Remove uncommented code here and below.

src/channel_request/channel_manager.rs
tnull: This is still unresolved. As this is *not* the same as the usual Lightning channel id, we should pick another name here.

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,834 @@
...
+	PendingConfirmation { request_id: RequestId, order_id: OrderId, invoice: Payment },
+
+	Ready { channel_id: u128, channel: CRChannel, channel_info: ChannelInfo },
+	// todo add abort/refund here
+}
+
+#[derive(Debug, PartialEq)]
+struct CRChannel {
+	channel_id: u128,
+	user_id: u128,
```
tnull: This isn't used anywhere and should be removed.

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,834 @@
...
+	counterparty_node_id: PublicKey,
+	state: ChannelStatus,
+	lsp_balance_sat: Option<u64>,
```
tnull: I assume this corresponds to the field on the order? Do we need to track this separately?
If so, should we maybe store the entire `Order` object here to have it separated out, but also to have access to more order fields if necessary?

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,834 @@
...
+	client_balance_sat: Option<u64>,
+	// confirms_within_blocks: Option<u32>,
+	// channel_expiry_blocks: Option<u32>,
+	announce_channel: bool,
+	order_id: Option<OrderId>,
```
tnull: See above, if we store the `Order` object that lead to the channel, this field would be superfluous.

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,834 @@
...
+	info: Option<ChannelInfo>,
```
tnull: I'm confused: what are the different use cases of `CRChannel` vs. `ChannelInfo`? Also, we store a `ChannelStatus` above but `ChannelInfo` (which isn't taken by reference here) holds another one?

Similarly, `CRState::Ready` holds `CRState` *and* yet another `ChannelInfo`, holding yet another `ChannelStatus`? 

All this seems very confusing and I doubt all these different state objects will be kept up-to-date without issue. Could you clarify your thinking behind these design choices?

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,834 @@
...
+	OM::Target: OnionMessageHandler,
+	CMH::Target: CustomMessageHandler,
+	NS::Target: NodeSigner,
+{
+	entropy_source: ES,
+	peer_manager: Mutex<Option<Arc<PeerManager<Descriptor, CM, RM, OM, L, CMH, NS>>>>,
+	pending_messages: Arc<Mutex<Vec<(PublicKey, LSPSMessage)>>>,
+	pending_events: Arc<Mutex<Vec<Event>>>,
+
+	per_peer_state: RwLock<HashMap<PublicKey, Mutex<CRState>>>,
```
tnull: Mh, this means we can only have one channel request per-peer at a time? What happens if we'd want to have multiple channels opened, or, if we want to keep some `CRState` around after the channel has been opened?

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,834 @@
...
+	pending_orders: RwLock<HashMap<OrderId, Order>>,
```
tnull: Are these inbound or outbound orders or both? Should we maybe split the LSP/client states internally, as otherwise this might get a bit confusing?

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,834 @@
...
+
+use super::msgs::{
+	ChannelInfo, ChannelStatus, CreateOrderResponse, GetInfoRequest, GetInfoResponse,
+	GetOrderRequest, GetOrderResponse, OrderId, OrderState, Payment, Response,
+};
+
+const SUPPORTED_SPEC_VERSION: u16 = 1;
+
+#[derive(Debug, PartialEq)]
+enum CRState {
```
tnull: Hm, let's at least create an `impl` here or on `CRChannel` that ensures we'd construct a new state via `new` and handle state transitions via method calls supplying the necessary info and taking `&mut self`. This way we can make sure that there are no invalid transitions and that we only transition if we have sufficient data. We could even further to introduce prettier state machine patterns, but this would at least be a start.

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,834 @@
...
+						counterparty_node_id
+					),
+					action: ErrorAction::IgnoreAndLog(Level::Info),
+				})
+			}
+		}
+		Ok(())
+	}
+
+	/* No errors for getinfo method */
```
tnull: nit: Please remove.

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,834 @@
...
+				}
+				super::msgs::Response::GetOrderError(error) => {
+					self.handle_get_order_error(request_id, counterparty_node_id, error)
+				}
+			},
+		}
+	}
+}
+
+// Order of functions called
```
tnull: As noted above, please remove all commented-out code and superfluous comments.

However, note that generally docs are missing. While we don't have that strong docs for non-public methods and objects, we still want to document things that we need to consider developing going forward. Would for example be great to clearly define what state represents what and which struct has which purpose exactly.

tnull (commented):
Did a high-level pass over the `channel_manager.rs` structures and found them a bit confusing tbh. Would be great to get some clarity here and as part of the docs.

src/channel_request/channel_manager.rs
```diff
@@ -29,35 +30,74 @@ use super::msgs::{
...
+	payment_invoice: OrderId,
+}
+
+struct ChannelOpen {
+	//channel: Payment,
+}
 
-	Ready { channel_id: u128, channel: CRChannel, channel_info: ChannelInfo },
-	// todo add abort/refund here
+enum ClientState {
```
tnull: I think the LSPS2 PR now used `InboundChannel`/`OutboundChannel` terminology, I think this would make sense here also, as the states are not really for a whole client or server, but per-channel, no?

src/channel_request/channel_manager.rs
```diff
@@ -29,35 +30,74 @@ use super::msgs::{
...
+struct ClientStateWrapper {
+	state: ClientState,
+}
+
+struct ServerStateWrapper {
+	state: ServerState,
+}
+
+struct PeerState {
+	pending_requests: HashMap<RequestId, ClientStateWrapper>,
```
tnull: Let's also here make clear in the field names whether they pertain `inbound` or `outbound` channel states.

src/channel_request/channel_manager.rs
```diff
@@ -29,35 +30,74 @@ use super::msgs::{
...
-	announce_channel: bool,
-	order_id: Option<OrderId>,
-	info: Option<ChannelInfo>,
+enum ServerState {
+	InfoResponse(InfoResponse),
+	OrderPlaced(OrderPlaced),
+	WaitingPayment(WaitingPayment),
+}
+
+struct ClientStateWrapper {
```
tnull: Why do we need these extra wrappers?

tnull (commented):
Excuse the delay here! I think now that #11 is getting refactored it would make sense to adjust this PR to follow the same model w.r.t. terminology and state (transition) handling.

Btw. seems to need a rebase (either on main or said #11) to get rid of the conflicts.

tnull:
Would love to land this first work-in-progress draft so that the code can be included in the upcoming project restructuring. For this, could you resolve any conflicts with current main and make sure it compiles and passes tests?

.github/workflows/build.yml
tnull, line 1: Please drop this commit, it's unrelated to LSPS1 and by now we should be upgrading to 0.0.118.

src/channel_request/msgs.rs
```diff
@@ -0,0 +1,192 @@
+use std::convert::TryFrom;
+
+use serde::{Deserialize, Serialize};
+
+//use bitcoin::hashes::hmac::{Hmac, HmacEngine};
```
tnull: Please cleanup any commented-out code.

src/channel_request/msgs.rs
```diff
@@ -0,0 +1,192 @@
...
+//use bitcoin::hashes::{Hash, HashEngine};
+use crate::transport::msgs::{LSPSMessage, RequestId, ResponseError};
+
+pub(crate) const LSPS1_GET_INFO_METHOD_NAME: &str = "lsps1.get_info";
+pub(crate) const LSPS1_CREATE_ORDER_METHOD_NAME: &str = "lsps1.create_order";
+pub(crate) const LSPS1_GET_ORDER_METHOD_NAME: &str = "lsps1.get_order";
+
+pub(crate) const REFUND_ONCHAIN_ADDRESS: bool = false;
+
+// Create a const to show preferred way for user payment
```
tnull: This comment seems to be a TODO/"note to self" and should be removed.

src/channel_request/msgs.rs
```diff
@@ -0,0 +1,192 @@
...
+	pub max_initial_lsp_balance_sat: u64,
+	pub min_channel_balance_sat: u64,
+	pub max_channel_balance_sat: u64,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct GetInfoResponse {
+	pub supported_versions: Vec<u16>,
+	pub website: String,
+	// This field should not be null
```
tnull: I'm not sure what this comment is trying to tell us?

src/channel_request/event.rs
```diff
@@ -0,0 +1,95 @@
+use bitcoin::secp256k1::PublicKey;
+
+use super::msgs::{ChannelInfo, OptionsSupported, Order, OrderId, Payment};
+use crate::transport::msgs::RequestId;
+
+/// An Event which you should probably take some action in response to.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Event {
+
```
tnull: Please remove the interspersed blank lines here and below.

src/channel_request/event.rs
```diff
@@ -0,0 +1,95 @@
+use bitcoin::secp256k1::PublicKey;
+
+use super::msgs::{ChannelInfo, OptionsSupported, Order, OrderId, Payment};
+use crate::transport::msgs::RequestId;
+
+/// An Event which you should probably take some action in response to.
```
tnull: nit: "event"  or "[`Event`]".

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,1000 @@
...
+	L::Target: Logger,
+	RM::Target: RoutingMessageHandler,
+	CM::Target: ChannelMessageHandler,
+	OM::Target: OnionMessageHandler,
+	CMH::Target: CustomMessageHandler,
+	NS::Target: NodeSigner,
+{
+	entropy_source: ES,
+	peer_manager: Mutex<Option<Arc<PeerManager<Descriptor, CM, RM, OM, L, CMH, NS>>>>,
+	channel_manager: Arc<ChannelManager<M, T, ES, NS, SP, F, R, L>>,
```
tnull, line 309: We can use the `AChannelManager` trait to at least make this a bit easier now.

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,1000 @@
...
+		version: u16,
+		order: Order,
+	},
+	PendingPayment { order_id: OrderId },
+	AwaitingConfirmation { id: u128, order_id: OrderId },
+}
+
+impl InboundRequestState {
+	fn info_received(&self, versions: Vec<u16>, options: OptionsSupported) -> Result<Self, ChannelStateError> {
+		let max_shared_version = versions
```
tnull, line 65: Note that we're in the process of dropping versioning support in all specs, so all the corresponding code can be removed: https://github.com/BitcoinAndLightningLayerSpecs/lsp/pull/63

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,1000 @@
...
+				"Received unexpected get_versions response. JIT Channel was in state: {:?}",
+				state
+			))),
+		}
+	}
+}
+
+struct OutboundCRChannelConfig {
+	options_supported: OptionsSupported,
+	/* supported_versions: Vec<u16>, */
```
tnull: nit: Please remove this comment.

src/channel_request/channel_manager.rs
```diff
@@ -0,0 +1,1000 @@
...
+		let onchain_payment = options_supported.min_onchain_payment_size_sat == None;
+		let supports_zero_channel_reserve = options_supported.supports_zero_channel_reserve;
+		
+		Self {
+			state: OutboundRequestState::OrderCreated { 
+				order_id
+			}, 
+			config: OutboundCRChannelConfig { 
+				options_supported, 
+				/* supported_versions: options_supported.supports_zero_channel_reserve,*/
```
tnull: nit: Like wise: please remove all commented-out code

src/transport/message_handler.rs
```diff
@@ -43,20 +45,38 @@ pub(crate) trait ProtocolMessageHandler {
...
 /// to provide liquidity services to clients.
-pub struct LiquidityProviderConfig;
+pub struct LiquidityProviderConfig {
+	/// LSPS1 Configuration
+	pub lsps1_config: Option<CRChannelConfig>,
+}
+
+pub struct CRChannelConfig {
+	pub token: Option<String>,
+
```
tnull: Please remove interspersed blank lines.

tnull (commented):
Thanks! 

Did a quick high-level pass. Will look more closely when this is rebased on main and the conflicts have been resolved. As stated above, let's do this soon so that all this code can be part of the restructuring PR coming up.

tnull:
@Srg213 Thank you for the rebase and addressing most of the comments! If you don't mind, could you undraft this PR? I'd then land it and start the restructuring work early next week. Any further changes to make this work could then happen as follow-ups.

tnull (approved):
Thanks! Going ahead and merge this draft so that we have a unified codebase. Any further changes can then be done in follow-up PRs.

PR #9, opened 2023-07-13
take and hold a reference to ChannelManager

johncantrell97:
There's no real need for the LiquidityManager to actually hold the channel manager.  Once LSPS1 and/or LSPS2 are landed we should remove the reference from LiquidityManager as the reference will just pass through to the constructors of the protocols.

Instead of putting the type constraints on just the `new` method, this way is closer to what it will actually look like once the `LiquidityManager` will hold LSPS1/2 objects that have these same generics.

### DIFF
--- a/src/transport/message_handler.rs
+++ b/src/transport/message_handler.rs
@@ -3,12 +3,16 @@ use crate::transport::msgs::{LSPSMessage, RawLSPSMessage, LSPS_MESSAGE_TYPE};
 use crate::transport::protocol::LSPS0MessageHandler;
 
 use bitcoin::secp256k1::PublicKey;
+use lightning::chain;
+use lightning::chain::chaininterface::{BroadcasterInterface, FeeEstimator};
+use lightning::ln::channelmanager::ChannelManager;
 use lightning::ln::features::{InitFeatures, NodeFeatures};
 use lightning::ln::msgs::{ErrorAction, LightningError};
 use lightning::ln::peer_handler::CustomMessageHandler;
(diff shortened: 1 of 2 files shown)


tnull:
The CI failure should be due to https://github.com/lightningdevkit/ldk-lsp-client/pull/10. Feel free to rebase this PR after it has been merged.

johncantrell97:
Rebased, thanks

src/transport/message_handler.rs
```diff
@@ -41,25 +45,53 @@ pub struct LiquidityProviderConfig;
 ///
 /// Should be used as a [`CustomMessageHandler`] for your
 /// [`lightning::ln::peer_handler::PeerManager`]'s [`lightning::ln::peer_handler::MessageHandler`].
-pub struct LiquidityManager<ES: Deref>
-where
+pub struct LiquidityManager<
+	ES: Deref,
```
tnull, line 49: @johncantrell97 Did you check that dropping all those `Send + Sync` bounds is indeed fine?

johncantrell97: nope, need to rebase lsps2 on top of this so I can check but wanted to make sure this idea was going to be accepted before rebasing

johncantrell97: ok just went through everything with c= node and we don't need any of the Send + Sync + 'static bounds at all.

tnull: Great!

tnull (approved):

PR #12, opened 2023-07-25
Add `Confirm`/`Listen`/`Filter` boilerplate

tnull:
As the LSPS implementations might need to have access to chain confirmation data (e.g., LSPS1/ChannelRequest would need to know when an on-chain transaction paying for a channel)

To this end, we add the necessary boilerplate to `LiquidityManager` for it to implement `Confirm`/`Listen`, take a `Filter` chain source, and then delegate the calls internally to the respective implementations.

Unfortunately the interfaces do not allow to watch for an arbitrary address getting paid, so we'll need to delegate this part via events to the on-chain wallet implemetation. After telling us about the txid of an inbound payment, we can then use `Filter`/`Confirm`/`Listen` to assert it is confirmed and isn't unconfirmed before we open a channel.


tnull:
Dropped the `FilterExt` as it doesn't make sense.

PR #15, opened 2023-07-27
Expose `lsps0_message_handler.handle_request`

yanganto:
Would `lsps0_message_handler.handle_request` be public and return the response?
Such that people integrate the handler will easier to verify the handler work after integration.
Thanks in advance. :pray:

### DIFF
--- a/src/transport/msgs.rs
+++ b/src/transport/msgs.rs
@@ -1,3 +1,4 @@
+#![allow(missing_docs)]
 use lightning::impl_writeable_msg;
 use lightning::ln::wire;
 use serde::de;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,3 +25,4 @@ mod transport;
 mod utils;
 
 pub use transport::message_handler::{LiquidityManager, LiquidityProviderConfig};
+pub use transport::msgs;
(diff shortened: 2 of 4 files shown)


src/transport/msgs.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(missing_docs)]
```
tnull, line 1: We set `#![deny(missing_docs)]` crate-wide for good reason and def. should allow them here.

tnull (commented):
I don't think we should expose arbitrary internals through the public API. So if you'll need access for testing I'd suggest to either change it locally or hide the exposure behind the `cfg(test)` flag.

src/transport/msgs.rs
yanganto: Is there demo client to request on the public interface of lsp handler?  It will be nice for people to check after integrate the handler.  Many thanks.

tnull: I think @johncantrell97 has a branch of `ldk-sample` with the LSP client and LSPS2 integrated. 

@johncantrell97, mind sharing a link here?

johncantrell97: I forget the state it's in but I think it was working: https://github.com/johncantrell97/ldk-sample/pull/1/files

PR #20, opened 2023-10-05
Jit channels clean

johncantrell97:
This implements LSPS2: JIT Channels on top of LSPS0 transport layer.


.github/workflows/build.yml
```diff
@@ -28,17 +28,15 @@ jobs:
...
           cargo update -p serde_json --precise "1.0.99" --verbose
           cargo update -p serde --precise "1.0.156" --verbose
           # The quote crate switched to Rust edition 2021 starting with v1.0.31, i.e., has MSRV of 1.56
           cargo update -p quote --precise "1.0.30" --verbose
           # The proc-macro2 crate switched to Rust edition 2021 starting with v1.0.66, i.e., has MSRV of 1.56
           cargo update -p proc-macro2 --precise "1.0.65" --verbose
-          # Sadly the log crate is always a dependency of tokio until 1.20, and has no reasonable MSRV guarantees
-          cargo update -p log --precise "0.4.18" --verbose
+          # The memchr crate switched to Rust edition 2021 starting with v2.6.0
+          cargo update -p memchr --precise "2.5.0" --verbose
```
tnull: Can we move the unrelated changes (e.g. pinning) to their PR?

johncantrell97: not following, do you mean to a separate PR?

tnull: Yes, sorry, meant to write "to their *own* PR".

johncantrell97: got it, ok

tnull:
This needs a rebase.

Cargo.toml
```diff
@@ -8,11 +8,10 @@ description = "Types and primitives to integrate a spec-compliant LSP with an LD
...
 
 [dependencies]
-lightning = { version = "0.0.116", features = ["max_level_trace", "std"] }
-lightning-invoice = "0.24.0"
-lightning-net-tokio = "0.0.116"
+lightning = { version = "0.0.117", default-features = false, features = ["max_level_trace", "no-std", "grind_signatures"] }
+lightning-invoice = { version = "0.25.0", default-features = false, features = ["no-std"] }
 
-bitcoin = "0.29.0"
+bitcoin = { version = "0.29.2", default-features = false, features = ["serde", "secp-recovery", "rand"] }
```
tnull: I think we may not need any of these features?

Cargo.toml
```diff
@@ -8,11 +8,10 @@ description = "Types and primitives to integrate a spec-compliant LSP with an LD
...
 
-serde = { version = "1.0", default-features = false, features = ["derive", "alloc"] }
+serde = { version = "^1.0", features = ["derive"] }
```
tnull: Why are we enabling default features? Also, the `^1.0` version is equivalent to `1.0`, would prefer to leave as is for consistency with other dependencies.

src/jit_channel/mod.rs
```diff
@@ -8,3 +8,4 @@
 // licenses.
 
 //! Types and primitives that implement the LSPS3: JIT Channel Negotiation specification.
```
tnull: `s/LSPS3/LSPS2/`

src/jit_channel/mod.rs
tnull, line 1: Let's expose this module in the crate root and avoid the direct imports of LSPS2-specific types in the crate root. E.g., `GetInfoResponse`, `JITChannelsConfig`, etc. are currently exposed in the root but should only live under this `jit_channel` module. Let's only expose stuff that makes sense in the context of *all* LSPSs in the crate root, e.g., `LiquidityManager`.

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+
+use crate::transport::msgs::{LSPSMessage, RequestId, ResponseError};
+use crate::utils;
+
+pub(crate) const LSPS2_GET_VERSIONS_METHOD_NAME: &str = "lsps2.get_versions";
+pub(crate) const LSPS2_GET_INFO_METHOD_NAME: &str = "lsps2.get_info";
+pub(crate) const LSPS2_BUY_METHOD_NAME: &str = "lsps2.buy";
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize, Default)]
+#[serde(default)]
```
tnull: I'm familiar with `serde(default)` for fields, but what does it do here for this struct and why do we need it?

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct GetInfoRequest {
+	pub version: u16,
+	pub token: Option<String>,
+}
+
+/// Fees and parameters for a JIT Channel.
+///
+/// The client will pay max(min_fee_msat, proportional*(payment_size_msat/1_000_000)).
```
tnull: Seems this is the wrong place to document this and we should document the fee structure (also) somewhere else in greater detail. But for the time being, can we tick the `max(..)` calculation here to have it at least set a bit nicer in the docs?

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+pub struct RawOpeningFeeParams {
+	/// The minimum fee required for the channel open.
+	pub min_fee_msat: u64,
+	/// A fee proportional to the size of the initial payment.
+	pub proportional: u32,
+	/// An ISO8601 formatted date for which these params are valid.
+	pub valid_until: String,
+	/// number of blocks that the LSP promises it will keep the channel alive without closing, after confirmation.
```
tnull: nit: `s/number/The number of blocks after confirmation`

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+	pub min_lifetime: u32,
+	/// Maximum number of blocks that the client is allowed to set its to_self_delay parameter.
```
tnull: nit: `s/Maximum/The maximum`, also tick `to_self_delay`

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+	pub max_client_to_self_delay: u32,
+}
+
+impl RawOpeningFeeParams {
+	pub(crate) fn into_opening_fee_params(self, promise_secret: &[u8; 32]) -> OpeningFeeParams {
```
tnull, line 59: Remind me, was there a reason we can't make this `impl From<RawOpeningFeeParams> for OpeningFeeParams`?

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+/// Fees and parameters for a JIT Channel.
+///
+/// The client will pay max(min_fee_msat, proportional*(payment_size_msat/1_000_000)).
+pub struct RawOpeningFeeParams {
+	/// The minimum fee required for the channel open.
+	pub min_fee_msat: u64,
+	/// A fee proportional to the size of the initial payment.
+	pub proportional: u32,
+	/// An ISO8601 formatted date for which these params are valid.
+	pub valid_until: String,
```
tnull: Wonder if it's worth taking a dependency on `chrono` and making this a `chrono::Datetime`? With the `serde` feature, this would also allow an ISO8601-compatible default implementation (see https://docs.rs/chrono/latest/chrono/serde/index.html / https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.to_rfc3339) ?

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+			valid_until: self.valid_until.clone(),
+			min_lifetime: self.min_lifetime,
+			max_client_to_self_delay: self.max_client_to_self_delay,
+			promise,
+		}
+	}
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+/// Fees and parameters for a JIT Channel with promise.
```
tnull: nit: The "with promise" seems a bit unclear here. We probably should dedicate a separate paragraph explaining what the promise is (for).

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+///
+/// The client will pay max(min_fee_msat, proportional*(payment_size_msat/1_000_000)).
```
tnull: Same as above: feels a bit out-of-place to mention the fee structure here, but we should at least tick the calculation to have it set properly in rendered docs.

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+
+/// Fees and parameters for a JIT Channel.
+///
+/// The client will pay max(min_fee_msat, proportional*(payment_size_msat/1_000_000)).
+pub struct RawOpeningFeeParams {
+	/// The minimum fee required for the channel open.
+	pub min_fee_msat: u64,
+	/// A fee proportional to the size of the initial payment.
+	pub proportional: u32,
+	/// An ISO8601 formatted date for which these params are valid.
```
tnull: Can we link to the ISO8601 standard here and below?

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+///
+/// The client will pay max(min_fee_msat, proportional*(payment_size_msat/1_000_000)).
+pub struct OpeningFeeParams {
+	/// The minimum fee required for the channel open.
+	pub min_fee_msat: u64,
+	/// A fee proportional to the size of the initial payment.
+	pub proportional: u32,
+	/// An ISO8601 formatted date for which these params are valid.
+	pub valid_until: String,
+	/// number of blocks that the LSP promises it will keep the channel alive without closing, after confirmation.
```
tnull: nit: `s/number of blocks/The number of blocks after confirmation`

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+	pub min_lifetime: u32,
+	/// Maximum number of blocks that the client is allowed to set its to_self_delay parameter.
```
tnull: nit: `s/Maximum/The maximum`, also tick `to_self_delay`

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+	pub max_client_to_self_delay: u32,
+	/// Field used by the LSP to validate that these parameters were actually given out by them.
```
tnull: This is a pretty unclear description, let's expand on this here or in the struct docs. Also "Field" is redundant here as all of these are fields.

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+	pub promise: String,
+}
+
+impl OpeningFeeParams {
+	/// Determine that these parameters are valid given the secret used to generate the promise.
```
tnull: nit: `s/Determine that these/Determines if the given/`

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+	// TODO: add validation check that valid_until >= now()
+	pub fn is_valid(&self, promise_secret: &[u8; 32]) -> bool {
```
tnull: It's a bit weird to have this function live here as part of the type because while the `promise: String` format is implementation-independent, this function is really specific to our particular implementation. Should this rather be a utility that takes the secret and `OpeningFeeParams`, or be a method on `JITChannelManager`?

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+	/// A set of opening fee parameters.
+	pub opening_fee_params_menu: Vec<OpeningFeeParams>,
+	/// The minimum payment size required to open a channel.
+	pub min_payment_size_msat: u64,
+	/// The maximum payment size the lsp will tolerate.
+	pub max_payment_size_msat: u64,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct BuyRequest {
```
tnull, line 113: This is missing docs.

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+		hmac.input(self.valid_until.as_bytes());
+		hmac.input(&self.min_lifetime.to_be_bytes());
+		hmac.input(&self.max_client_to_self_delay.to_be_bytes());
+		let promise_bytes = Hmac::from_engine(hmac).into_inner();
+		let promise = utils::hex_str(&promise_bytes[..]);
+		promise == self.promise
+	}
+}
+
+/// Information about the parameters a LSP is willing to offer clients
```
tnull: nit: Maybe just "A response to a [`GetInfoRequest`]"?

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct BuyRequest {
+	pub version: u16,
+	pub opening_fee_params: OpeningFeeParams,
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub payment_size_msat: Option<u64>,
+}
+
+/// A response from a buy request made by a client
```
tnull: nit: "A response to a [`BuyRequest`]".

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+///
+/// Includes information needed to construct an invoice.
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct BuyResponse {
+	/// The short channel id used by LSP to identify need to open channel.
+	pub jit_channel_scid: String,
```
tnull: Should we introduce a `JitChannelScid` type for this that allows easy conversion between this string format and `u64`, which is used LDK-wide?

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+	/// The locktime expiry delta the lsp requires.
+	pub lsp_cltv_expiry_delta: u32,
+	/// A flag that indicates who is trusting who.
+	#[serde(default)]
```
tnull, line 154: Why do we need this `default`?

src/transport/message_handler.rs
```diff
@@ -154,6 +153,15 @@ where {
 			LSPSMessage::LSPS0(msg) => {
 				self.lsps0_message_handler.handle_message(msg, sender_node_id)?;
 			}
+			_ => {
+				return Err(LightningError {
+					err: format!(
+						"Received message without message handler configured. From node = {:?}",
```
tnull: nit: Probably could be "From node {}"?

src/transport/msgs.rs
```diff
@@ -20,22 +23,25 @@ const JSONRPC_INVALID_MESSAGE_ERROR_CODE: i32 = -32700;
 const JSONRPC_INVALID_MESSAGE_ERROR_MESSAGE: &str = "parse error";
 const LSPS0_LISTPROTOCOLS_METHOD_NAME: &str = "lsps0.list_protocols";
 
-pub const LSPS_MESSAGE_TYPE: u16 = 37913;
+/// The Lightning message type id for LSPS messages
+pub const LSPS_MESSAGE_TYPE_ID: u16 = 37913;
 
+/// Lightning message type used by LSPS protocols
```
tnull: nit: We should decide whether to end lines with a `.` to keep things consistent. If so, we should add it here and elsewhere.

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,263 @@
...
+pub enum Response {
+	GetVersions(GetVersionsResponse),
+	GetInfo(GetInfoResponse),
+	GetInfoError(ResponseError),
+	Buy(BuyResponse),
+	BuyError(ResponseError),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Message {
```
tnull: To keep things consistent with `LSPS0`, we may want to rename this `LSPS2Message`, and the `Request`/`Response` types to `LSPS2Request`/`LSPS2Response` accordingly. At least for now. If we ever would reconsider we should change the naming scheme overall, but let's first make it consistent at least.

src/transport/msgs.rs
```diff
@@ -248,6 +320,52 @@ impl<'de, 'a> Visitor<'de> for LSPSMessageVisitor<'a> {
...
+					jit_channel::msgs::LSPS2_GET_VERSIONS_METHOD_NAME => {
+						if let Some(result) = result {
+							let response =
+								serde_json::from_value(result).map_err(de::Error::custom)?;
+							Ok(LSPSMessage::LSPS2(jit_channel::msgs::Message::Response(
+								RequestId(id),
+								jit_channel::msgs::Response::GetVersions(response),
+							)))
+						} else {
+							Err(de::Error::custom("Received invalid lsps2.getversions response."))
```
tnull: `s/getversions/get_versions/`

src/events.rs
```diff
@@ -53,6 +53,9 @@ impl EventQueue {
 	}
 }
 
-/// Event which you should probably take some action in response to.
+/// An event which you should probably take some action in response to.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum Event {}
+pub enum Event {
+	/// A LSPS2 (JIT Channel) protocol event
```
tnull: nit: "An LSPS2.."

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
+// This file is Copyright its original authors, visible in version contror
```
tnull: Seems we included a typo in all of the license headers. Please `s/contror/control/` everywhere and double-check nothing else has been changed.

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+use bitcoin::secp256k1::PublicKey;
+
+use super::msgs::OpeningFeeParams;
+use crate::transport::msgs::RequestId;
+
+/// An Event which you should probably take some action in response to.
```
tnull: nit: "An event" as above.

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Event {
```
tnull: Should this be called `LSPS2Event`?

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+	/// A request from a client for information about JIT Channel parameters.
+	///
+	/// You must calculate the paramaeters for this client and pass them to
```
tnull: `s/paramaeters/parameters/`

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+	/// [`crate::LiquidityManager::opening_fee_params_generated`].
```
tnull: nit: Let's avoid the `crate` prefix when linking to other parts of the docs and rather introduce a 

```
	/// [`LiquidityManager::opening_fee_params_generated`]: crate::LiquidityManager::opening_fee_params_generated
```

line at the end (here and everywhere applicable).

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+	GetInfo {
+		/// An identifier that must be passed to [`crate::LiquidityManager::opening_fee_params_generated`].
+		request_id: RequestId,
+		/// The node id of the client making the information request.
+		counterparty_node_id: PublicKey,
+		/// The protocol version they would like to use.
+		version: u16,
+		/// An optional token that can be used as an api key, coupon code, etc.
```
tnull: nit: `s/api/API/`

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+		token: Option<String>,
+	},
+	/// Information from LSP about their current fee and channel parameters.
```
tnull: nit: "from the LSP"

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+	///
+	/// You must call [`crate::LiquidityManager::opening_fee_params_selected`] with the fee parameter
+	/// you want to use if you wish to proceed opening a channel.
+	GetInfoResponse {
+		/// Needs to be passed to [`crate::LiquidityManager::opening_fee_params_selected`].
```
tnull: Let's first state that this is a channel id, possibly we also want to clarify here that this is *not* *the* channel id, but yet another id for the channel (request)..

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+		min_payment_size_msat: u64,
+		/// The max payment size allowed when opening the channel.
+		max_payment_size_msat: u64,
+		/// The user_channel_id value passed in to [`crate::LiquidityManager::create_invoice`].
+		user_channel_id: u128,
+	},
+	/// A client has selected a opening fee parameter to use and would like to
+	/// purchase a channel with an optional initial payment size.
+	///
+	/// If payment_size_msat is [`Option::Some`] then the payer is allowed to use MPP
```
tnull: nit: Tick `payment_size_msat`

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+	/// If payment_size_msat is [`Option::None`] then the payer cannot use MPP
+	///
+	/// You must generate an scid and cltv_expiry_delta for them to use
```
tnull: nit: Tick `cltv_expiry_delta`

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+	InvoiceGenerationReady {
+		/// The node id of the LSP.
+		counterparty_node_id: PublicKey,
+		/// The short channel id to use in the route hint.
+		scid: u64,
+		/// The cltv_expiry_delta to use in the route hint.
+		cltv_expiry_delta: u32,
+		/// The initial payment size you specified.
+		payment_size_msat: Option<u64>,
+		/// The trust model the lsp expects.
```
tnull: nit: `s/lsp/LSP` here and elsewhere.

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+	/// Use the provided fields to generate an invoice and give to payer.
+	///
+	/// When the invoice is paid the LSP will open a channel to you
+	/// with the previously agreed upon parameters.
+	InvoiceGenerationReady {
+		/// The node id of the LSP.
+		counterparty_node_id: PublicKey,
+		/// The short channel id to use in the route hint.
+		scid: u64,
+		/// The cltv_expiry_delta to use in the route hint.
```
tnull: nit: Tick

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+		/// The max payment size allowed when opening the channel.
+		max_payment_size_msat: u64,
+		/// The user_channel_id value passed in to [`crate::LiquidityManager::create_invoice`].
+		user_channel_id: u128,
+	},
+	/// A client has selected a opening fee parameter to use and would like to
+	/// purchase a channel with an optional initial payment size.
+	///
+	/// If payment_size_msat is [`Option::Some`] then the payer is allowed to use MPP
+	/// If payment_size_msat is [`Option::None`] then the payer cannot use MPP
```
tnull: nit: Tick

src/jit_channel/event.rs
```diff
@@ -0,0 +1,104 @@
...
+		counterparty_node_id: PublicKey,
+		/// The short channel id to use in the route hint.
+		scid: u64,
+		/// The cltv_expiry_delta to use in the route hint.
+		cltv_expiry_delta: u32,
+		/// The initial payment size you specified.
+		payment_size_msat: Option<u64>,
+		/// The trust model the lsp expects.
+		client_trusts_lsp: bool,
+		/// The user_channel_id value passed in to [`crate::LiquidityManager::create_invoice`].
```
tnull: nit: Tick

src/lib.rs
```diff
@@ -24,4 +24,5 @@ mod jit_channel;
 mod transport;
 mod utils;
 
+pub use jit_channel::event::Event as JITChannelEvent;
```
tnull: As mentioned above: let's not expose this in the crate root, but keep it under `jit_channel`.

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+	L::Target: Logger,
+	RM::Target: RoutingMessageHandler,
+	CM::Target: ChannelMessageHandler,
+	OM::Target: OnionMessageHandler,
+	CMH::Target: CustomMessageHandler,
+	NS::Target: NodeSigner,
+{
+	entropy_source: ES,
+	peer_manager: Mutex<Option<Arc<PeerManager<Descriptor, CM, RM, OM, L, CMH, NS>>>>,
+	channel_manager: Arc<ChannelManager<M, T, ES, NS, SP, F, R, L>>,
```
tnull, line 377: Since LDK 0.0.117, we can use the `AChannelManager` trait here now.

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+};
+use lightning::ln::peer_handler::{CustomMessageHandler, PeerManager, SocketDescriptor};
+use lightning::ln::ChannelId;
+use lightning::routing::router::Router;
+use lightning::sign::{EntropySource, NodeSigner, SignerProvider};
+use lightning::util::errors::APIError;
+use lightning::util::logger::{Level, Logger};
+
+use crate::events::EventQueue;
+use crate::jit_channel::msgs::Message;
```
tnull: nit: Lets move this down into the other `msgs` import.

src/jit_channel/scid_utils.rs
tnull, line 1: Probably this should live under `crate::util` as it's not JIT channel specific?

tnull (commented):
Started a thorough pass, still have to take a closer look at message handler commit and following.

src/jit_channel/scid_utils.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/// Extracts the vout (bytes [0..2]) from the `short_channel_id`
+pub fn vout_from_scid(short_channel_id: &u64) -> u16 {
+	((short_channel_id) & MAX_SCID_VOUT_INDEX) as u16
+}
+
+pub fn scid_from_human_readable_string(
+	human_readable_scid: &str,
+) -> Result<u64, ShortChannelIdError> {
```
tnull: If `InvalidScid` will be the only error variant, we could also just return `Err(())` instead?

Also, see https://github.com/lightningdevkit/ldk-lsp-client/pull/20#discussion_r1361750272

src/utils.rs
```diff
@@ -67,3 +67,13 @@ pub fn parse_pubkey(pubkey_str: &str) -> Result<PublicKey, std::io::Error> {
 
 	Ok(pubkey.unwrap())
 }
+
+pub fn compute_opening_fee(
```
tnull: I think this *is* LSPS2 specific, so we need to move it to `jit_channel/utils.rs`? Also it could use some docs, and we possibly want to expose it?

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+	pub payment_size_msat: Option<u64>,
+}
+
+#[derive(PartialEq, Debug)]
+enum InboundJITChannelState {
+	VersionsRequested,
+	MenuRequested,
+	PendingMenuSelection,
+	BuyRequested,
+	PendingPayment { client_trusts_lsp: bool, scid: String },
```
tnull: nit: `s/scid/short_channel_id/` for consistency.

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+	) -> Result<(), LightningError> {
+		self.state = self.state.invoice_params_received(client_trusts_lsp, jit_channel_scid)?;
+		Ok(())
+	}
+}
+
+#[derive(PartialEq, Debug)]
+enum OutboundJITChannelState {
+	InvoiceParametersGenerated {
+		scid: u64,
```
tnull: nit: `s/scid/short_channel_id/` for consistency.

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+	) -> Result<Self, ChannelStateError> {
+		match self {
+			OutboundJITChannelState::InvoiceParametersGenerated { opening_fee_params, .. } => {
+				let opening_fee_msat: Option<u64> = utils::compute_opening_fee(
+					expected_outbound_amount_msat,
+					opening_fee_params.min_fee_msat,
+					opening_fee_params.proportional as u64,
+				);
+
+				if let Some(opening_fee_msat) = opening_fee_msat {
```
tnull: nit: Making this an `unwrap_or_else` would probably make things a bit more streamlined.

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+
+	pub fn htlc_intercepted(
+		&self, expected_outbound_amount_msat: u64, intercept_id: InterceptId,
+	) -> Result<Self, ChannelStateError> {
+		match self {
+			OutboundJITChannelState::InvoiceParametersGenerated { opening_fee_params, .. } => {
+				let opening_fee_msat: Option<u64> = utils::compute_opening_fee(
+					expected_outbound_amount_msat,
+					opening_fee_params.min_fee_msat,
+					opening_fee_params.proportional as u64,
```
tnull: Can avoid the conversion (here) if we have `compute_opening_fee` take `proportional: u32`.

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+			pending_messages,
+			pending_events,
+			per_peer_state: RwLock::new(HashMap::new()),
+			peer_by_scid: RwLock::new(HashMap::new()),
+			peer_manager: Mutex::new(None),
+			channel_manager,
+		}
+	}
+
+	fn map_scid_to_peer(&self, scid: u64, counterparty_node_id: PublicKey) {
```
tnull: This method name is kinda misleading to me, at least I wouldn't expect that it inserts anything in our peer map. As it's just two lines, we may want to inline them, which possibly would also improve our ability to reason about lock handling?

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+		opening_fee_params: OpeningFeeParams,
+	) -> Result<(), APIError> {
+		let outer_state_lock = self.per_peer_state.read().unwrap();
+		match outer_state_lock.get(&counterparty_node_id) {
+			Some(inner_state_lock) => {
+				let mut peer_state = inner_state_lock.lock().unwrap();
+				if let Some(jit_channel) =
+					peer_state.inbound_channels_by_id.get_mut(&jit_channel_id)
+				{
+					if let Err(e) = jit_channel.opening_fee_params_selected() {
```
tnull: Could this be a simple `map_err(..)?`?

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+										expected_outbound_amount_msat,
+										amt_to_forward_msat,
+										opening_fee_msat,
+										user_channel_id: scid as u128,
+									},
+								));
+							}
+							Err(e) => {
+								self.channel_manager.fail_intercepted_htlc(intercept_id)?;
+								// remove channel?
```
tnull: Probably?

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+	}
+
+	fn enqueue_event(&self, event: Event) {
+		self.pending_events.enqueue(event);
+	}
+
+	fn handle_get_versions_request(
+		&self, request_id: RequestId, counterparty_node_id: &PublicKey,
+	) -> Result<(), LightningError> {
+		// not sure best way to extract a vec to a constant? lazy_static?
```
tnull: Probably need to make `SUPPORTED_SPEC_VERSIONS` an array and convert it to `Vec` here.

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+						),
+						action: ErrorAction::IgnoreAndLog(Level::Info),
+					})?;
+
+				if let Err(e) = jit_channel.info_received() {
+					peer_state.remove_inbound_channel(jit_channel_id);
+					return Err(e);
+				}
+
+				self.enqueue_event(Event::LSPS2(super::event::Event::GetInfoResponse {
```
tnull: Let's import the (renamed) `LSPS2Event` rather than using `super` for disambiguation.

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+				let jit_channel_id =
+					peer_state.request_to_cid.remove(&request_id).ok_or(LightningError {
+						err: format!(
+							"Received get_info error for an unknown request: {:?}",
+							request_id
+						),
+						action: ErrorAction::IgnoreAndLog(Level::Info),
+					})?;
+
+				let _jit_channel = peer_state
```
tnull: nit: No need for this `let`?

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1119 @@
...
+					.get_mut(&jit_channel_id)
+					.ok_or(LightningError {
+						err: format!(
+							"Received buy response for an unknown channel: {:?}",
+							jit_channel_id
+						),
+						action: ErrorAction::IgnoreAndLog(Level::Info),
+					})?;
+
+				if let Err(e) = jit_channel.invoice_params_received(
```
tnull, line 1099: nit: Same as above, this can probably be an `unwrap_or_else`?

src/transport/message_handler.rs
```diff
@@ -42,20 +45,48 @@ pub(crate) trait ProtocolMessageHandler {
 ///
 /// Allows end-user to configure options when using the [`LiquidityManager`]
 /// to provide liquidity services to clients.
-pub struct LiquidityProviderConfig;
+pub struct LiquidityProviderConfig {
+	/// Optional configuration for jit channels
```
tnull: nit: `s/jit/JIT/` here and elsewhere in docs.

src/transport/message_handler.rs
```diff
@@ -42,20 +45,48 @@ pub(crate) trait ProtocolMessageHandler {
...
+	/// should you want to support them
+	pub jit_channels: Option<JITChannelsConfig>,
+}
+
+/// Configuration options for jit channels
+/// A configuration used for the creation of Just In Time Channels.
+pub struct JITChannelsConfig {
```
tnull, line 59: No requirement right now, but do we eventually want to enable setting some default parameters here rather than supplying all in-line via events?

src/transport/message_handler.rs
```diff
@@ -42,20 +45,48 @@ pub(crate) trait ProtocolMessageHandler {
...
 /// [`lightning::ln::peer_handler::PeerManager`]'s [`lightning::ln::peer_handler::MessageHandler`].
+///
+/// Should provide a reference to your [`lightning::ln::peer_handler::PeerManager`] by calling
+/// [`LiquidityManager::set_peer_manager()`] post construction.  This allows the [`LiquidityManager`] to
+/// wake the [`lightning::ln::peer_handler::PeerManager`] when there are pending messages to be sent.
+///
+/// Users need to continually poll [`LiquidityManager::get_and_clear_pending_events()`] in order to surface
+/// [`Event`]'s that likely need to be handled.
+///
+/// Users must forward the [`lightning::events::Event::HTLCIntercepted`] event parameters to [`LiquidityManager::htlc_intercepted()`]
```
tnull: nit: `LiquidityManager::htlc_intercepted`, always leave out `()`.

src/transport/message_handler.rs
```diff
@@ -99,10 +141,14 @@ where
...
 	L::Target: Logger,
+	RM::Target: RoutingMessageHandler,
+	CM::Target: ChannelMessageHandler,
+	OM::Target: OnionMessageHandler,
+	CMH::Target: CustomMessageHandler,
 	NS::Target: NodeSigner,
 	C::Target: Filter,
 {
-	/// Constructor for the LiquidityManager
+	/// Constructor for the ['LiquidityManager']
```
tnull: nit: Wrong ticks.

src/transport/message_handler.rs
```diff
@@ -203,6 +207,165 @@ where {
 		self.pending_events.get_and_clear_pending_events()
 	}
 
+	/// Set a [`lightning::ln::peer_handler::PeerManager`] reference for the message handlers
```
tnull: As mentioned before, let's link to this as [`PeerManager`] and provide a fully qualified link below:

```
[`PeerManager`]: lightning::ln::peer_handler::PeerManager
```

(also follow this pattern elsewhere)

src/transport/message_handler.rs
```diff
@@ -203,6 +207,165 @@ where {
...
+			lsps2_message_handler.set_peer_manager(peer_manager);
+		}
+	}
+
+	/// Initiate the creation of an invoice that when paid will open a channel
+	/// with enough inbound liquidity to be able to receive the payment.
+	///
+	/// `counterparty_node_id` is the node_id of the LSP you would like to use.
+	///
+	/// if `payment_size_msat` is [`Option::Some`] then the invoice will be for a fixed amount
```
tnull: nit: `s/if/If/`. Also end sentences with `.`, here and elsewhere.

src/transport/message_handler.rs
```diff
@@ -203,6 +207,165 @@ where {
...
+	/// and MPP can be used to pay it.
+	///
+	/// if `payment_size_msat` is [`Option::None`] then the invoice can be for an arbitrary amount
+	/// but MPP can no longer be used to pay it.
+	///
+	/// `token` is an optional String that will be provided to the LSP.
+	/// it can be used by the LSP as an API key, coupon code, or some other way to identify a user.
+	pub fn create_invoice(
```
tnull: For now, let's prefix the LSPS2-specific methods in the API with `lsps2` or `jit_channel` to avoid confusion with other LSPSs.

src/transport/message_handler.rs
```diff
@@ -213,15 +376,14 @@ where {
 			LSPSMessage::LSPS0(msg) => {
 				self.lsps0_message_handler.handle_message(msg, sender_node_id)?;
 			}
-			_ => {
```
tnull: Do we still need this catchall though?

tnull (commented):
Completed the pass. 

Looks pretty good to me, but I have yet to do another pass comparing the logic once more to the spec.

Cargo.toml
johncantrell97: yeah probably not, I made all these changes to dependencies when trying to get Mutiny integration working. will remove the changes for now

Cargo.toml
johncantrell97: same reason as above (mutiny) -- will figure out what is actually needed to support their browser wasm compilation and enable it via feature over here instead of being the default?

src/jit_channel/msgs.rs
johncantrell97: hm, maybe we don't actually need it. I think I just copied this from the serde(default) on the ListProtocolsRequest {} in LSPS0.

I think I thought we needed it in order to deserialize the JSON-RPC string when/if it is missing params.  params is optional in JSON-RPC.

I have a test for this in transport::msgs and it passes if I remove the serde(default).  

I'll just remove these for now.

src/jit_channel/msgs.rs
johncantrell97: yes probably, I don't feel like implementing this and dealing with the expiration is something that is still TBD so haven't properly felt the pain yet.

using `chrono` should simplify a lot of the work there.

src/jit_channel/event.rs
johncantrell97: probably. originally I thought you'd just use jit_channel::Event everywhere but it's a bit confusing.  A separate name seems simpler and clearer.

src/jit_channel/event.rs
johncantrell97: renamed to `jit_channel_id` and added more context.

src/jit_channel/msgs.rs
johncantrell97: switched to using chrono

src/jit_channel/msgs.rs
johncantrell97: Because it needs user to provide the promise_secret, it's not stored in RawOpeningFeeParams. 

The reason we have two of these objects is that we want the LSP to be able to generate their fee params without needing to know anything about the existence of this promise stuff.

To go from the "Raw" to the one that gets sent over the wire we need to use the secret to calculate the promise.

src/jit_channel/msgs.rs
johncantrell97: yeah, actually I think OpeningFeeParams (with promise) should not be exposed at all.  We should probably have a `From<OpeningFeeParams> for RawOpeningFeeParams` where we strip out the promise before surfacing it to the end-user to select a fee.

johncantrell97: I guess the challenge of doing that is that we actually need to send back the fee param with promise to the LSP when we issue the BuyRequest.  This means we would have to keep around the menu in the InboundJitChannelState and match it to the one the end-user picked as part of `opening_fee_params_selected`.

 Not a huge deal.. think it's worth it to avoid having to document and expose the `promise` concept?

src/jit_channel/msgs.rs
johncantrell97: agreed, I think we should probably just get rid of this being public (see comment above).

src/jit_channel/msgs.rs
johncantrell97: makes sense, moved it to a util

src/jit_channel/msgs.rs
johncantrell97: The spec says this parameter is optional but default to false which is the behavior we get with serde(default).

src/jit_channel/msgs.rs
johncantrell97: ok, im fine with that. I think the reason I didn't do it (similar to Event earlier) is I thought it would all be namespaced by jit_channel::

src/transport/message_handler.rs
johncantrell97: yeah, for LSP side might want to include `min_payment_size_msat` and `max_payment_size_msat`.  Though I wonder if we'll just need to provide overrides anyway.

for client side maybe we let them pick a 'strategy' for how they want to select a fee from the menu automatically and not surface opening fee params at all.  Allow them to set min/max fee ranges and/or target lifetime 🤷  Seems like it might get complicated but I don't imagine a lot of wallets (or LSPs for that matter) surfacing a fee menu to users.

src/transport/message_handler.rs
johncantrell97: I'd prefer `create_jit_channel_invoice`?

tnull: Yes, I'd prefer that too, but we should try to keep the scheme ~consistent across all such methods.

src/jit_channel/msgs.rs
tnull: Mh, but would this mean we'd always *write* it? But maybe that's exactly what we want, even if technically we could omit it?

src/jit_channel/msgs.rs
tnull: > Because it needs user to provide the promise_secret, it's not stored in RawOpeningFeeParams.

Good point, not sure what I was thinking. :grin:

src/jit_channel/msgs.rs
tnull: Nice! Once this lands I'll have to look into no-std stuff, which might conflict with the `chrono` approach, but we'll figure it out when we get to it. Generally it would be preferable if we could use it.

src/jit_channel/msgs.rs
tnull: I'd say we could consider expose all basic message types as they are mentioned in the spec just in case someone else wants to reuse them somehow. So exposing `promise` would be fine with me here, as long as we spend a sentence documenting what it means.

src/jit_channel/channel_manager.rs
johncantrell97: is this better than just changing opening_fee_params to hold a u64?

src/jit_channel/channel_manager.rs
johncantrell97: hm seems like Id want `.ok_or(..)?` to convert the Option<u64> to Result<u64, E> and use ? to unwrap or early return the error.

`unwrap_or_else` would require me to provide a default `opening_fee_msat` which doesn't really make sense here.

johncantrell97: could also throw in a map there and just go:

`compute_opening_fee(...).map(|fee| ...).ok_or(ChannelStateError(...))`

src/jit_channel/channel_manager.rs
johncantrell97: yeah I guess I did it because we only need the lock to insert then it should be dropped.  That means when written inline I would typically nest it in it's own { } block so the lock is dropped whereas extracting to separate method does that for me and the method name kind of acts as a comment to describe what's happening.

fine inlining it though 🤷

src/jit_channel/channel_manager.rs
johncantrell97: yeah it could, is it weird to do the cleanup `peer_state.remove_inbound_channel(...)` as part of the map_err() though?

johncantrell97: hm tried it and also run into issue borrowing `peer_state` twice since it needs to get moved into the closure

src/jit_channel/channel_manager.rs
johncantrell97: hm i guess technically if it failed because of invalid fee computation the payer could send again and re-use scid?
do we also remove scid from the peer_by_scid map? seems like we'd need to restructure this fn a bit in order to not run into borrow checker issues here and do the removal at the end/outside of these nested if/matches

src/jit_channel/channel_manager.rs
johncantrell97: ok this was a bit of mess but probably needed at some point.  Converted it to an array and then had to fix version negotiation.

Now we take the maximum shared version between client and LSP and choose that as the InboundChannel version.  It's stored in the channel state and used when making subsequent requests to the LSP for that channel.

src/jit_channel/channel_manager.rs
johncantrell97: hm not quite because we want to early return on error. i can't really move it into a map_err(...)? either because of the borrowing peer_state mutably twice.

src/jit_channel/msgs.rs
johncantrell97: I think we would always write it because it's not possible to construct a BuyResponse without setting it.  It's also currently required for the LSP to provide a value when calling `invoice_parameters_generated`.

I guess we could probably just make it an Option<bool> and whenever it's surfaced to end-user unwrap_or_default() it.  I assume that would allow us to not always write it (set it to None if user passes in false).

I guess I don't really feel strongly, though I'd lean towards not changing it at this point.  It's a bit odd it's optional in the spec, i assume it's just to save a few bytes in the payload?

src/jit_channel/msgs.rs
johncantrell97: Ok sounds good. With regards to original comment, I think since the spec has OpeningFeeParams with the promise included, it's a bit weird to call it out explicitly.  we can still describe it here (and maybe link the spec?).

The "RawOpeningFeeParams"  should be described as OpeningFeeParams without promise.  I don't know if "Raw" is really a good adjective to use either.

src/jit_channel/msgs.rs
johncantrell97: agreed, moved it to where client selects the fees they agree to use when calling`opening_fee_params_selected`

src/jit_channel/msgs.rs
johncantrell97: moved it to where client agrees to an opening fee when calling `opening_fee_params_selected`

src/transport/msgs.rs
johncantrell97: went with `.`

src/transport/message_handler.rs
johncantrell97: no because all patterns are matched? i think this was an artifact of me splitting up the commits.

src/jit_channel/msgs.rs
tnull: Yeah, I agree we can just always write it.

src/jit_channel/channel_manager.rs
tnull: Yeah, that might be a solution. Up to you, when in question we can also just leave as is. We just already have many `if`s and `match`es and reducing them and the indendation levels might help with readability going forward.

src/jit_channel/channel_manager.rs
tnull: Alright, fine by me to leave as-is then.

src/jit_channel/channel_manager.rs
tnull: Alright, fine to leave as-is then for now.

src/jit_channel/channel_manager.rs
tnull: Okay thank you. Probably not worth touching it again until the spec group decided which of the 3 versioning proposals we'll go forward with.

src/jit_channel/channel_manager.rs
tnull: Yeah, let's inline it and just use a dedicated block if we want to drop the lock early.

src/jit_channel/channel_manager.rs
tnull: Mh, I also considered that, but the spec explicitly makes a point of it fitting in a `u32`...

johncantrell97: ah ok, I already made the change you suggested

src/jit_channel/channel_manager.rs
```diff
@@ -0,0 +1,1239 @@
...
+					);
+					return Err(LightningError {
+						err: "overflow error when calculating opening_fee".to_string(),
+						action: ErrorAction::IgnoreAndLog(Level::Info),
+					});
+				}
+			}
+		}
+
+		// TODO: if payment_size_msat is specified, make sure our node has sufficient incoming liquidity from public network to receive it.
```
johncantrell97, line 1035: This is mentioned in the spec but doesn't really feel like it belongs in the library.  I think we probably need to expose a failure flow for the LSP to call into as an alternative to `invoice_parameters_generated` when they receive a BuyRequest event.

This way if they decide they can't honor/fulfill the request for whatever reason they have a way to bail.  We would add this spec comment to the description of BuyRequest Event for a potential reason they might fail it.

tnull: Yeah, that makes sense. LSP-side liquidity management is really out-of-scope of this library.

src/jit_channel/channel_manager.rs
```diff
@@ -377,6 +380,8 @@ pub struct JITChannelManager<
 	per_peer_state: RwLock<HashMap<PublicKey, Mutex<PeerState>>>,
 	peer_by_scid: RwLock<HashMap<u64, PublicKey>>,
 	promise_secret: [u8; 32],
+	min_payment_size_msat: u64,
```
tnull, line 383: Will probably need all of the config options, so might be worth holding a `JitChannelConfig` here?

src/jit_channel/msgs.rs
```diff
@@ -0,0 +1,324 @@
...
+	/// An [`ISO8601`](https://www.iso.org/iso-8601-date-and-time-format.html) formatted date for which these params are valid.
+	pub valid_until: chrono::DateTime<Utc>,
+	/// The number of blocks after confirmation that the LSP promises it will keep the channel alive without closing.
+	pub min_lifetime: u32,
+	/// T maximum number of blocks that the client is allowed to set its `to_self_delay` parameter.
+	pub max_client_to_self_delay: u32,
+}
+
+impl RawOpeningFeeParams {
+	pub(crate) fn into_opening_fee_params(self, promise_secret: &[u8; 32]) -> OpeningFeeParams {
```
tnull, line 59: Still would be nice to have this be a util method analogous to `is_valid_opening_fee` (could also DRY this one up then) to keep LSP-specific logic separate from the messaging/type serialization.

Happy to have that happen in a follow-up though.

Cargo.toml
```diff
@@ -1,18 +1,18 @@
...
 [dependencies]
-lightning = { version = "0.0.116", features = ["max_level_trace", "std"] }
-lightning-invoice = "0.24.0"
-lightning-net-tokio = "0.0.116"
+lightning = { version = "0.0.117", default-features = false, features = ["max_level_trace", "std"] }
+lightning-invoice = "0.25.0"
 
 bitcoin = "0.29.0"
 
+chrono = { version = "0.4", default-features = false, features = ["std", "serde"] }
```
tnull, line 16: Seems we don't actually require full `std` here, `alloc` should be enough.

tnull: Now tracking here: https://github.com/lightningdevkit/ldk-lsp-client/issues/29#issuecomment-1784994836

src/jit_channel/msgs.rs
tnull: Now tracking here: https://github.com/lightningdevkit/ldk-lsp-client/issues/30

src/transport/message_handler.rs
```diff
@@ -43,20 +49,56 @@ pub(crate) trait ProtocolMessageHandler {
...
 /// to provide liquidity services to clients.
-pub struct LiquidityProviderConfig;
+pub struct LiquidityProviderConfig {
+	/// Optional configuration for JIT channels
+	/// should you want to support them.
+	pub jit_channels: Option<JITChannelsConfig>,
+}
+
+/// Configuration options for JIT channels.
+pub struct JITChannelsConfig {
```
tnull, line 59: This probably should be renamed `OutboundJitChannelConfig` or `LSPS2OutboundChannelConfig` and get moved to the `jit_channel`/`lsps2` module.

tnull: Now tracking here https://github.com/lightningdevkit/ldk-lsp-client/issues/31

tnull (approved):
Alright. I had another look at the spec and poked around a bit more. 

While nothing in particular stood out, we'll def. want want to increase our input validation, error handling and test coverage going forward, i.e., before this is deemed stable and usable in production. 

I had some minor comments which however can be addressed as follow ups. To this end, and to generally track next steps, I now opened a bunch of issues.

That said, for now, I'm merging this first draft :tada:

PR #22, opened 2023-10-24
Add MPP support to LSPS2

johncantrell97:
This should be based on top of lightningdevkit/lightning-liquidity#20 but I can't open a PR on top of a PR in my repo over here.  So if reviewing this please just look at the last commit.  Once lightningdevkit/lightning-liquidity#20 is merged I will rebase this PR.

When a payment is made that initiates a JIT channel open it is possible the payer uses MPP. If `payment_size_msat` is specified by client then the LSP needs to allow for MPP to be used.

This means when an HTLC is intercepted we need to collect the payment parts and only open the channel once the total amount we collected is enough to cover the original payment_size.

This PR introduces a Vec of `InterceptedHTLC`'s into the OutboundJITChannelState to track these.  There's a new `OutboundJITChannelState::PendingPaymentReceived` state that can be entered if MPP is supported and the first htlc_intercepted is not large enough to cover the requested payment_size.

It is possible to skip this state if MPP is not supported or the first payment is large enough.

The spec does not specify how the LSP should extract the fee from the collection of HTLCs received as part of the MPP.  In this pass I iterate over all parts and greedily take as much of the fee I can from each HTLC until the total fee amount has been accounted for.  I could imagine another strategy that tries to minimize the number of HTLCs that are used to pay the fee (sorting from largest to smallest and performing same strategy but I'm not sure it matters.

It's also unclear to me whether batch forwarding all of the HTLCs at once could run into problems in practice.

### DIFF
--- a/src/jit_channel/channel_manager.rs
+++ b/src/jit_channel/channel_manager.rs
@@ -13,7 +13,6 @@ use std::ops::Deref;
 use std::sync::{Arc, Mutex, RwLock};
 
 use bitcoin::secp256k1::PublicKey;
-use lightning::chain;
 use lightning::chain::chaininterface::{BroadcasterInterface, FeeEstimator};
 use lightning::ln::channelmanager::{ChannelManager, InterceptId};
 use lightning::ln::msgs::{
@@ -25,6 +24,7 @@ use lightning::routing::router::Router;
 use lightning::sign::{EntropySource, NodeSigner, SignerProvider};
 use lightning::util::errors::APIError;
 use lightning::util::logger::{Level, Logger};
+use lightning::{chain, log_debug};
 
 use crate::events::EventQueue;
 use crate::jit_channel::utils::{compute_opening_fee, is_valid_opening_fee_params};
@@ -192,17 +192,11 @@ impl InboundJITChannel {
 
 #[derive(PartialEq, Debug)]
 enum OutboundJITChannelState {
-	InvoiceParametersGenerated {
-		short_channel_id: u64,
-		cltv_expiry_delta: u32,
-		payment_size_msat: Option<u64>,
-		opening_fee_params: OpeningFeeParams,
-	},
-	PendingPaymentReceived {
+	AwaitingPayment {
+		min_fee_msat: u64,
+		proportional_fee: u32,
 		htlcs: Vec<InterceptedHTLC>,
-		payment_size_msat: u64,
-		amt_to_forward_msat: u64,
-		opening_fee_msat: u64,
+		payment_size_msat: Option<u64>,
 	},
 	PendingChannelOpen {
 		htlcs: Vec<InterceptedHTLC>,
@@ -216,33 +210,49 @@ enum OutboundJITChannelState {
 }
 
 impl OutboundJITChannelState {
-	pub fn new(
-		short_channel_id: u64, cltv_expiry_delta: u32, payment_size_msat: Option<u64>,
-		opening_fee_params: OpeningFeeParams,
-	) -> Self {
-		OutboundJITChannelState::InvoiceParametersGenerated {
-			short_channel_id,
-			cltv_expiry_delta,
+	pub fn new(payment_size_msat: Option<u64>, opening_fee_params: OpeningFeeParams) -> Self {
+		OutboundJITChannelState::AwaitingPayment {
+			min_fee_msat: opening_fee_params.min_fee_msat,
+			proportional_fee: opening_fee_params.proportional,
+			htlcs: vec![],
 			payment_size_msat,
-			opening_fee_params,
 		}
 	}
 
 	pub fn htlc_intercepted(&self, htlc: InterceptedHTLC) -> Result<Self, ChannelStateError> {
 		match self {
-			OutboundJITChannelState::InvoiceParametersGenerated {
-				opening_fee_params,
+			OutboundJITChannelState::AwaitingPayment {
+				htlcs,
 				payment_size_msat,
-				..
+				min_fee_msat,
+				proportional_fee,
 			} => {
-				let htlcs = vec![htlc];
-				let supports_mpp = payment_size_msat.is_some();
-				let payment_size_msat =
-					payment_size_msat.unwrap_or(htlc.expected_outbound_amount_msat);
-				let opening_fee_msat = compute_opening_fee(payment_size_msat, opening_fee_params.min_fee_msat, opening_fee_params.proportional.into()).ok_or(ChannelStateError(format!("Could not compute valid opening fee with min_fee_msat = {}, proportional = {}, and expected_outbound_amount_msat = {}", opening_fee_params.min_fee_msat, opening_fee_params.proportional, htlc.expected_outbound_amount_msat)))?;
-				let amt_to_forward_msat = payment_size_msat - opening_fee_msat;
+				let mut htlcs = htlcs.clone();
+				htlcs.push(htlc);
 
-				if htlc.expected_outbound_amount_msat >= payment_size_msat
+				let total_expected_outbound_amount_msat =
+					htlcs.iter().map(|htlc| htlc.expected_outbound_amount_msat).sum();
+
+				let supports_mpp = payment_size_msat.is_some();
+
+				let expected_payment_size_msat =
+					payment_size_msat.unwrap_or(total_expected_outbound_amount_msat);
+
+				let opening_fee_msat = compute_opening_fee(
+					expected_payment_size_msat,
+					*min_fee_msat,
+					(*proportional_fee).into(),
+				).ok_or(ChannelStateError(
+					format!("Could not compute valid opening fee with min_fee_msat = {}, proportional = {}, and total_expected_outbound_amount_msat = {}", 
+						min_fee_msat,
+						proportional_fee,
+						total_expected_outbound_amount_msat
+					)
+				))?;
+
+				let amt_to_forward_msat = expected_payment_size_msat - opening_fee_msat;
+
+				if total_expected_outbound_amount_msat >= expected_payment_size_msat
 					&& amt_to_forward_msat > 0
 				{
 					Ok(OutboundJITChannelState::PendingChannelOpen {
@@ -252,44 +262,16 @@ impl OutboundJITChannelState {
 					})
 				} else {
 					if supports_mpp {
-						Ok(OutboundJITChannelState::PendingPaymentReceived {
+						Ok(OutboundJITChannelState::AwaitingPayment {
+							min_fee_msat: *min_fee_msat,
+							proportional_fee: *proportional_fee,
 							htlcs,
-							amt_to_forward_msat,
-							opening_fee_msat,
-							payment_size_msat,
+							payment_size_msat: *payment_size_msat,
 						})
 					} else {
 						Err(ChannelStateError("HTLC is too small to pay opening fee".to_string()))
 					}
 				}
-			}
-			OutboundJITChannelState::PendingPaymentReceived {
-				htlcs,
-				payment_size_msat,
-				amt_to_forward_msat,
-				opening_fee_msat,
-			} => {
-				let mut htlcs = htlcs.clone();
-				htlcs.push(htlc);
-
-				let total_expected_outbound_amount_msat = htlcs
-					.iter()
-					.fold(0, |total_msat, htlc| total_msat + htlc.expected_outbound_amount_msat);
-
-				if total_expected_outbound_amount_msat >= *payment_size_msat {
-					return Ok(OutboundJITChannelState::PendingPaymentReceived {
-						htlcs,
-						payment_size_msat: *payment_size_msat,
-						amt_to_forward_msat: *amt_to_forward_msat,
-						opening_fee_msat: *opening_fee_msat,
-					});
-				}
-
-				Ok(OutboundJITChannelState::PendingChannelOpen {
-					htlcs,
-					opening_fee_msat: *opening_fee_msat,
-					amt_to_forward_msat: *amt_to_forward_msat,
-				})
 			}
 			state => Err(ChannelStateError(format!(
 				"Invoice params received when JIT Channel was in state: {:?}",
@@ -316,20 +298,21 @@ impl OutboundJITChannelState {
 
 struct OutboundJITChannel {
 	state: OutboundJITChannelState,
+	scid: u64,
+	cltv_expiry_delta: u32,
+	client_trusts_lsp: bool,
 }
 
 impl OutboundJITChannel {
 	pub fn new(
-		scid: u64, cltv_expiry_delta: u32, payment_size_msat: Option<u64>,
+		scid: u64, cltv_expiry_delta: u32, client_trusts_lsp: bool, payment_size_msat: Option<u64>,
 		opening_fee_params: OpeningFeeParams,
 	) -> Self {
 		Self {
-			state: OutboundJITChannelState::new(
-				scid,
-				cltv_expiry_delta,
-				payment_size_msat,
-				opening_fee_params,
-			),
+			scid,
+			cltv_expiry_delta,
+			client_trusts_lsp,
+			state: OutboundJITChannelState::new(payment_size_msat, opening_fee_params),
 		}
 	}
 
@@ -339,7 +322,10 @@ impl OutboundJITChannel {
 		self.state = self.state.htlc_intercepted(htlc)?;
 
 		match &self.state {
-			OutboundJITChannelState::PendingPaymentReceived { .. } => Ok(None),
+			OutboundJITChannelState::AwaitingPayment { htlcs, payment_size_msat, .. } => {
+				// log that we received an htlc but are still awaiting payment
+				Ok(None)
+			}
 			OutboundJITChannelState::PendingChannelOpen {
 				opening_fee_msat,
 				amt_to_forward_msat,
@@ -644,6 +630,7 @@ where
 						let outbound_jit_channel = OutboundJITChannel::new(
 							scid,
 							cltv_expiry_delta,
+							client_trusts_lsp,
 							buy_request.payment_size_msat,
 							buy_request.opening_fee_params,
 						);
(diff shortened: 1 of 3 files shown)


src/jit_channel/channel_manager.rs
```diff
@@ -216,20 +228,68 @@ impl OutboundJITChannelState {
...
+			} => {
+				let htlcs = vec![htlc];
+				let supports_mpp = payment_size_msat.is_some();
+				let payment_size_msat =
+					payment_size_msat.unwrap_or(htlc.expected_outbound_amount_msat);
+				let opening_fee_msat = compute_opening_fee(payment_size_msat, opening_fee_params.min_fee_msat, opening_fee_params.proportional).ok_or(ChannelStateError(format!("Could not compute valid opening fee with min_fee_msat = {}, proportional = {}, and expected_outbound_amount_msat = {}", opening_fee_params.min_fee_msat, opening_fee_params.proportional, htlc.expected_outbound_amount_msat)))?;
+				let amt_to_forward_msat = payment_size_msat - opening_fee_msat;
+
+				if htlc.expected_outbound_amount_msat >= payment_size_msat
+					&& amt_to_forward_msat > 0
```
johncantrell97, line 255: This should probably be something like `amt_to_forward_msat >= min_htlc_msat` but there's no channel yet so we don't know what the minimum htlc amount our peer will set on the channel.  I don't think the spec mentions anything about this.

tnull:
This needs rebase now that lightningdevkit/lightning-liquidity#20 was merged.

src/jit_channel/channel_manager.rs
```diff
@@ -192,13 +198,19 @@ enum OutboundJITChannelState {
 		payment_size_msat: Option<u64>,
 		opening_fee_params: OpeningFeeParams,
 	},
+	PendingPaymentReceived {
```
tnull: The name is a bit ambiguous as it could be read in the sense that we *did* receive the pending payment, which is exactly the opposite of what the state is for. Maybe `PendingPaymentReceipt`?

src/jit_channel/channel_manager.rs
```diff
@@ -216,20 +228,68 @@ impl OutboundJITChannelState {
...
+			OutboundJITChannelState::InvoiceParametersGenerated {
+				opening_fee_params,
+				payment_size_msat,
+				..
+			} => {
+				let htlcs = vec![htlc];
+				let supports_mpp = payment_size_msat.is_some();
+				let payment_size_msat =
+					payment_size_msat.unwrap_or(htlc.expected_outbound_amount_msat);
+				let opening_fee_msat = compute_opening_fee(payment_size_msat, opening_fee_params.min_fee_msat, opening_fee_params.proportional).ok_or(ChannelStateError(format!("Could not compute valid opening fee with min_fee_msat = {}, proportional = {}, and expected_outbound_amount_msat = {}", opening_fee_params.min_fee_msat, opening_fee_params.proportional, htlc.expected_outbound_amount_msat)))?;
```
tnull: This line is more than 400 characters wide. Can we please break it up?

src/jit_channel/channel_manager.rs
```diff
@@ -216,20 +228,68 @@ impl OutboundJITChannelState {
...
+				payment_size_msat,
+				amt_to_forward_msat,
+				opening_fee_msat,
+			} => {
+				let mut htlcs = htlcs.clone();
+				htlcs.push(htlc);
+
+				let total_expected_outbound_amount_msat = htlcs
+					.iter()
+					.fold(0, |total_msat, htlc| total_msat + htlc.expected_outbound_amount_msat);
```
tnull: nit: Could write this a tad simpler:
```suggestion
					.map(|htlc| htlc.expected_outbound_amount_msat).sum();
```

src/jit_channel/channel_manager.rs
```diff
@@ -216,20 +228,68 @@ impl OutboundJITChannelState {
...
+
+				if total_expected_outbound_amount_msat >= *payment_size_msat {
```
tnull: This is the wrong way around, no? If our total HTLC values are still *smaller* than the payment size, we remain  in `PendingPaymentReceived`?

src/jit_channel/channel_manager.rs
```diff
@@ -216,20 +228,68 @@ impl OutboundJITChannelState {
...
+							amt_to_forward_msat,
+							opening_fee_msat,
+							payment_size_msat,
+						})
+					} else {
+						Err(ChannelStateError("HTLC is too small to pay opening fee".to_string()))
+					}
+				}
+			}
+			OutboundJITChannelState::PendingPaymentReceived {
```
tnull: I wonder whether it would make sense to have this be in the default path, even when we don't support MPP? We could have our logic discern between MPP and non-MPP requests and just always check the necessary parameters in this state? If they are met, we advance to `PendingChannelOpen`, if not, we timeout after a while?

src/jit_channel/channel_manager.rs
```diff
@@ -625,25 +683,17 @@ where
 				Some(inner_state_lock) => {
 					let mut peer_state = inner_state_lock.lock().unwrap();
 					if let Some(jit_channel) = peer_state.outbound_channels_by_scid.get_mut(&scid) {
-						// TODO: Need to support MPP payments. If payment_amount_msat is known, needs to queue intercepted HTLCs in a map by payment_hash
-						//       LiquidityManager will need to be regularly polled so it can continually check if the payment amount has been received
-						//       and can release the payment or if the channel valid_until has expired and should be failed.
```
tnull, line 630: I take it you don't intend to handle timeouts/`valid_until` in this PR?

src/jit_channel/channel_manager.rs
```diff
@@ -625,25 +683,17 @@ where
...
 								self.enqueue_event(Event::LSPS2(LSPS2Event::OpenChannel {
 									their_network_key: counterparty_node_id.clone(),
-									inbound_amount_msat,
-									expected_outbound_amount_msat,
 									amt_to_forward_msat,
 									opening_fee_msat,
 									user_channel_id: scid as u128,
 								}));
 							}
+							Ok(None) => {}
```
tnull, line 683: Should we at least log our progress here?

src/jit_channel/channel_manager.rs
```diff
@@ -675,13 +725,39 @@ where
...
+									.iter()
+									.find(|channel| channel.channel_id == *channel_id)
+									.map(|details| details.next_outbound_htlc_minimum_msat)
+									.ok_or(APIError::APIMisuseError {
+										err: format!("Channel with id {} not found", channel_id),
+									})?;
+
+								let total_msat = htlcs
+									.iter()
+									.fold(0, |acc, htlc| acc + htlc.expected_outbound_amount_msat);
```
tnull: Same nit as above: lets use `sum` for this.

src/jit_channel/channel_manager.rs
```diff
@@ -675,13 +725,39 @@ where
...
+								let mut fee_msat = total_msat - amt_to_forward_msat;
+
+								for htlc in htlcs {
+									let max_we_can_take =
+										htlc.expected_outbound_amount_msat - min_htlc_msat;
+									let amount_of_fee_to_take =
+										std::cmp::min(fee_msat, max_we_can_take);
+									let amount_to_forward_msat =
```
tnull: Let's not shadow `amt_to_forward_msat`, this is confusing.

src/jit_channel/channel_manager.rs
```diff
@@ -675,13 +725,39 @@ where
...
+									})?;
+
+								let total_msat = htlcs
+									.iter()
+									.fold(0, |acc, htlc| acc + htlc.expected_outbound_amount_msat);
+								let mut fee_msat = total_msat - amt_to_forward_msat;
+
+								for htlc in htlcs {
+									let max_we_can_take =
+										htlc.expected_outbound_amount_msat - min_htlc_msat;
```
tnull: I believe `next_outbound_htlc_minimum_msat` might change over time and would be influenced by `forward_intercepted_htlc`, so taking *exactly* this amount without updating would likely fail.

src/jit_channel/channel_manager.rs
```diff
@@ -675,13 +725,39 @@ where
...
+									let amount_of_fee_to_take =
+										std::cmp::min(fee_msat, max_we_can_take);
```
tnull: I'm not quite getting why we'd take all fees from the first HTLCs but leave all the others untouched? Shouldn't we rather split it equally over all?

src/jit_channel/channel_manager.rs
johncantrell97: Yeah good point.  I think even can/should get rid of the first state of InvoiceParametersGenerated because really soon as that is given out you are already awaiting payment with an empty vec of htlcs received.

johncantrell97: well the state is really supposed to be that we received _part_ of the payment.  it only transitions here once the first htlc is received.

i wonder though if we should just get rid of the InvoiceParametersGenerated state and just immediately enter some kind of AwaitingPayment state that holds the htlcs and initializes with empty vec.

johncantrell97: maybe it's beneficial to immediately split into one of two states whether or not MPP is supported (based on presence of the payment_size_msat).

This probably simplifies the logic when an htlc arrives. It allows for the fast failure of non-mpp supported payments that are < required amount.

tnull: Yeah, that could be a solution. I'll leave it up to you. I also commented something similar regarding simplifying the states below. Here my main comment was that the naming is ambiguous and we should at least rename the state.

src/jit_channel/channel_manager.rs
johncantrell97: Easier to read but probably _slightly_ less efficient unless the compiler does some magic to avoid two loops? I'll go with easier to read regardless because the # of htlcs is tiny.

src/jit_channel/channel_manager.rs
johncantrell97: yes seems backwards

src/jit_channel/channel_manager.rs
johncantrell97: Yeah, but we have issues logged in the repo for the timeouts so I felt it was safe to remove in-code "todo comments".

src/jit_channel/channel_manager.rs
johncantrell97: yes probably. we currently don't have access to a `Logger` in channel state or even the JitChannelManager right now.  Should we pass one around everywhere here?

src/jit_channel/channel_manager.rs
johncantrell97: yeah makes sense.  i could move the fetching of `next_outbound_htlc_minimum_msat` from the channel into the loop I guess so it happens before each htlc.

src/jit_channel/channel_manager.rs
johncantrell97: well it's not necessarily just from the first:

in the PR I wrote:


> The spec does not specify how the LSP should extract the fee from the collection of HTLCs received as part of the MPP. In this pass I iterate over all parts and greedily take as much of the fee I can from each HTLC until the total fee amount has been accounted for. I could imagine another strategy that tries to minimize the number of HTLCs that are used to pay the fee (sorting from largest to smallest and performing same strategy but I'm not sure it matters.


We can't split the fee msat amount equally across all htlcs in case some htlcs are too small.  I guess we could try to do it proportionally from each HTLC? So calculate `fee_percentage = fee_msat / total_msat` and then take `htlc_amount_msat*fee_percentage` for each one? Feels like that could run into rounding issues and be slightly off.

I'm open to whatever strategy, the spec doesn't mention how to do it.

src/jit_channel/channel_manager.rs
johncantrell97: hm it's not _really_ shadowing it. I think I'll rename `amt_to_forward_msat` to `total_amt_to_forward_msat` and keep this one.

src/jit_channel/channel_manager.rs
johncantrell97: I ended up removing both states and now the initial state of an outbound channel is `AwaitingPayment`. If mpp is enabled it stays there until receives enough to cover payment_size+fee.  If mpp not enabled then after the first htlc is received, if it's not enough it fails immediately.

johncantrell97:
I just pushed a new commit that handles _most_ of your review comments.  I simplified and renamed the OutboundJITChannelState to have the initial state immediately be `AwaitingPayment`.  

I have a couple of questions still:

1) Should I have JITChannelManager and the channel state objects take/hold a `Logger`? Right now they have no way to log outside of returning error messages that will get logged upstream.

2) The strategy for taking the fee from htlcs.  Currently I greedily take as much as I can from each htlc until I've collected enough for the total fee.  You suggested splitting it evenly, which would require doing so proportionally as a percentage of each htlc amount.  I think that would work but might run into rounding issues.

johncantrell97:
1) We decided to punt on logging for now until the restructuring is settled.
2) I pushed a commit that takes the fee proportionally from each HTLC.  It rounds up to nearest msat for each HTLC and then takes the min(fee_remaining, proportional_fee) so that the last HTLC doesn't take more than whatever is left.  If we didn't take the min then the user would slightly overpay on fees (~1msat per htlc, roughly).

src/jit_channel/channel_manager.rs
```diff
@@ -25,6 +24,7 @@ use lightning::routing::router::Router;
 use lightning::sign::{EntropySource, NodeSigner, SignerProvider};
 use lightning::util::errors::APIError;
 use lightning::util::logger::{Level, Logger};
+use lightning::{chain, log_debug};
```
tnull: nit: As we'll punt on logging, we shouldn't import the logging macro here.

src/jit_channel/channel_manager.rs
```diff
@@ -186,50 +192,86 @@ impl InboundJITChannel {
...
+					*proportional_fee,
+				).ok_or(ChannelStateError(
+					format!("Could not compute valid opening fee with min_fee_msat = {}, proportional = {}, and total_expected_outbound_amount_msat = {}", 
+						min_fee_msat,
+						proportional_fee,
+						total_expected_outbound_amount_msat
+					)
+				))?;
+
+				let amt_to_forward_msat = expected_payment_size_msat - opening_fee_msat;
```
tnull: Are we sure this can't underflow? Can we use `saturating_sub` here?

src/jit_channel/channel_manager.rs
```diff
@@ -186,50 +192,86 @@ impl InboundJITChannel {
...
+				if total_expected_outbound_amount_msat >= expected_payment_size_msat
+					&& amt_to_forward_msat > 0
+				{
+					Ok(OutboundJITChannelState::PendingChannelOpen {
+						htlcs,
+						opening_fee_msat,
+						amt_to_forward_msat,
+					})
+				} else {
+					if supports_mpp {
```
tnull: Can we just make this an `if let Some(_)` and include a comment? Or at least move the `supports_mpp` down, as its defined way before we need it.

src/jit_channel/channel_manager.rs
```diff
@@ -258,34 +298,39 @@ impl OutboundJITChannelState {
...
-		&mut self, expected_outbound_amount_msat: u64, intercept_id: InterceptId,
-	) -> Result<(u64, u64), LightningError> {
-		self.state = self.state.htlc_intercepted(expected_outbound_amount_msat, intercept_id)?;
+		&mut self, htlc: InterceptedHTLC,
+	) -> Result<Option<(u64, u64)>, LightningError> {
+		self.state = self.state.htlc_intercepted(htlc)?;
 
 		match &self.state {
+			OutboundJITChannelState::AwaitingPayment { htlcs, payment_size_msat, .. } => {
+				// log that we received an htlc but are still awaiting payment
```
tnull: Let's include a `TODO` here so that we may find it again:
```suggestion
				// TODO: Should we log that we received an HTLC but are still awaiting payment?
```

src/jit_channel/channel_manager.rs
```diff
@@ -675,13 +712,40 @@ where
...
+								// TODO: review strategy for handling batch forwarding of these htlcs
+								let total_received_msat: u64 = htlcs
+									.iter()
+									.map(|htlc| htlc.expected_outbound_amount_msat)
+									.sum();
+
+								let mut fee_remaining_msat =
+									total_received_msat - total_amt_to_forward_msat;
+								let fee_percentage =
+									fee_remaining_msat as f64 / total_received_msat as f64;
```
tnull: Let's not introduce `f64`s here, they just make it hard to deal with fractional `msat`s and rounding errors.

src/jit_channel/channel_manager.rs
```diff
@@ -675,13 +712,40 @@ where
...
+										htlc.expected_outbound_amount_msat - actual_fee_amt_msat;
+
+									self.channel_manager.forward_intercepted_htlc(
+										htlc.intercept_id,
+										channel_id,
+										*counterparty_node_id,
+										amount_to_forward_msat,
+									)?;
+
+									fee_remaining_msat -= actual_fee_amt_msat;
```
tnull: Are we sure we're not introducing some error here? How much would the LSP get underpaid here? Can we pull this logic out into a dedicated method and run some unit tests on it?

src/jit_channel/channel_manager.rs
johncantrell97: Yeah either way I guess. I was trying to make it clear that Some(payment_size_msat) is how we determine if we support_mpp or not.

I think originally I was shadowing payment_size_msat so had moved the supports_mpp calculation up before unwrapping + shadowing it.  Since that's not the case I can move it down.

src/jit_channel/channel_manager.rs
johncantrell97: Sounds good, should probably create an issue too?

src/jit_channel/channel_manager.rs
johncantrell97: hm, how do you avoid it when working with percentages? i tried to avoid rounding errors below by always rounding up to nearest msat and then taking min of remaining amount

src/jit_channel/channel_manager.rs
johncantrell97: hm? they can't get underpaid? I can still extract it and add tests to convince you :)

johncantrell97: ok pulled it out and added a test. not sure what other scenario to test exactly.

src/jit_channel/channel_manager.rs
```diff
@@ -1237,3 +1281,70 @@ where
 		}
 	}
 }
+
+fn calculate_amount_to_forward(
```
tnull: nit: Can we include 'htlc' or 'per-htlc' in the method name to make it a bit clearer what it does?

src/jit_channel/channel_manager.rs
```diff
@@ -1237,3 +1281,70 @@ where
...
+}
+
+#[cfg(test)]
+mod tests {
+
+	use super::*;
+
+	#[test]
+	fn test_calculate_amount_to_forward() {
+		let htlcs = vec![
```
tnull, line 1325: Could you add a TODO here saying `// TODO: Use proptest to generate random allocations` so we would be reminded to come back here?

Would be a good first test case to start implementing property-based testing. Happy to do that as a follow-up, tracking here: https://github.com/lightningdevkit/ldk-lsp-client/issues/45

src/jit_channel/channel_manager.rs
```diff
@@ -1237,3 +1281,70 @@ where
...
+
+fn calculate_amount_to_forward(
+	htlcs: &[InterceptedHTLC], total_amt_to_forward_msat: u64,
+) -> Vec<(InterceptId, u64)> {
+	let total_received_msat: u64 =
+		htlcs.iter().map(|htlc| htlc.expected_outbound_amount_msat).sum();
+
+	let mut fee_remaining_msat = total_received_msat - total_amt_to_forward_msat;
+
+	let fee_percentage = fee_remaining_msat as f64 / total_received_msat as f64;
```
tnull: Still not convinced we really want/need these casting to `f64`.

tnull (commented):
Generally looks good to me, mod some nits and the known/documented issues we want to explore further.

src/jit_channel/channel_manager.rs
johncantrell97: ok, I pushed a version using integer division but I'm way less confident in it.

tnull (approved):
LGTM.

Landing this in preparation of the restructuring to spare you the rebasing. Will also add some further test coverage later today.

PR #41, opened 2023-11-15
Update LSPS2's `compute_opening_fee`

tnull:
While the intial Rust code proposed in the spec wasn't wrong, it has since been updated to a more idiomatic version. Here, we update our methods accordingly.

### DIFF
--- a/src/jit_channel/channel_manager.rs
+++ b/src/jit_channel/channel_manager.rs
@@ -224,7 +224,7 @@ impl OutboundJITChannelState {
 				compute_opening_fee(
 					expected_outbound_amount_msat,
 					opening_fee_params.min_fee_msat,
-					opening_fee_params.proportional,
+					opening_fee_params.proportional.into(),
 				).map(|opening_fee_msat| OutboundJITChannelState::PendingChannelOpen {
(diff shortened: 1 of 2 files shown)


src/jit_channel/channel_manager.rs
```diff
@@ -224,7 +224,7 @@ impl OutboundJITChannelState {
 				compute_opening_fee(
 					expected_outbound_amount_msat,
 					opening_fee_params.min_fee_msat,
-					opening_fee_params.proportional,
+					opening_fee_params.proportional.into(),
```
johncantrell97, line 227: should we just store proportional as a u64 in [Raw]OpeningFeeParams and avoid these conversions?

tnull: I think we discussed this before, I have no strong opinion here. I'd however rather not touch the messages in this PR, so if thats fine by you we could just land this and then change it later if we really want to do so?

johncantrell97: yeah i thought we had talked about this before. okay sounds good

johncantrell97 (approved):

PR #46, opened 2023-11-27
allow for optional chain parameters

johncantrell97:
LSPS2 users currently do not need to track the chain and therefore integration is slightly complicated by requiring ChainParameters.

Until LSPS2 also requires chain tracking we should make this optional.


tnull (approved):
Yeah, makes sense. We should eventually document what implementations require what parameters, but that should probably be part of an interface rework.

Landing this for now.

tnull:
Ah, wait, CI is failing due to missing `cargo fmt` (@johncantrell97)

johncantrell97:
Thanks, fixed formatting.

johncantrell97:
@tnull thanks, fixed formatting.

Issue #1, opened 2023-04-05
Project structure / Naming conventions


tnull:
For now I created the following submodules in the crate:

- `transport` for LSPS0
- `channel_request` for LSPS1
- `jit_channel` for LSPS3 (possibly LSPS2 soon)

I however wonder if we should follow the namespacing introduced in the LSPSs and structure the crate along the lines of 
`crate::lsps0::..` etc.

A question arising from this would then how to deal with datatypes shared between LSPSs...

tnull:
It seems current sentiment in the spec group is is to *not* have  LSPS4 rename to LSPS2v2. It's probably therefore best to rename and restructure this crate along the lines of the LSPS numbers, i.e., have the modules named:
`lsps0`, `lsps1`, `lsps2`, etc.

Moreover, we should really rename the crate as a whole. Given we might want to upstream it eventually, we could consider already renaming it to `lightning-liquidity-management` or `lightning-liquidity-tools`.

I'd likely want to do this once #20  lands, and before we cut an intial alpha release.
(cc @johncantrell97)

johncantrell97:
sounds good, can do the renaming and restructuring post merge.  I agree the name of the crate is horrible.  I think those suggestions are already much better but might be too broad? 

I could imagine there being other liquidity management tools (swaps, liquidity ads, etc.) that might not belong in this crate? I guess they could end up in here?

tnull:
> I could imagine there being other liquidity management tools (swaps, liquidity ads, etc.) that might not belong in this crate? I guess they could end up in here?

Yeah, that's a good point. Given that we call the main object `LiquidityManager`, we might actually consider whether it could support these other `Liquidity` things if they were added eventually?

jkczyz:
> Moreover, we should really rename the crate as a whole. Given we might want to upstream it eventually, we could consider already renaming it to `lightning-liquidity-management` or `lightning-liquidity-tools`.

I'd opt for something short like `lightning-liquidity`. If you use a longer name, someone can come around later and take the shorter one and get more eyeballs / feel more canonical.

tnull:
[...]

PR #48, opened 2023-11-28
Restructure crate

### LINKED ISSUE #34: Project Restructuring: Move LSPS implementations to `lspsX` modules
As discussed out-of-band, we'll move to a `crate::lspsX` module format afterall.

We should hence restructure the project accordingly once #20 (and possibly follow-ups) lands.

tnull:
Fixes #34 

The structure of this crate was up for discussion right from the beginning (see #1) and we tentatively chose a semantic structure with modules named after their purpose (`transport`, `channel_request`, `jit_channel`). 

However, given that the specifications now follow a "no versioning" model in which breaking changes will result in a new LSPSX standard this structure doesn't really make sense anymore. In particular when LSPS4 will become a reality it live side-by-side with LSPS2, so we will have two `jit_channel` standards for a while. 

To this end, we rename the modules accordingly and `cfg`-gate LSPS1 for now as it's not usable yet, although now checked in CI to make sure we won't break compatibility with it.


src/lsps2/channel_manager.rs
```diff
@@ -662,21 +659,21 @@ where
 							peer_by_scid.insert(scid, counterparty_node_id);
 						}
 
-						let outbound_jit_channel = OutboundJITChannel::new(
+						let outbound_lsps2 = OutboundJITChannel::new(
```
johncantrell97: hm this feels awkward, also OutboundJITChannel still exists.  Maybe we can keep jit around in some cases as long as it's prefixed by lsps2?

let outbound_lsps2_jit_channel = OutboundLSPS2JITChannel::new()

or let outbound_lsps2_channel = ...

src/lsps2/channel_manager.rs
```diff
@@ -704,9 +701,9 @@ where
 			match outer_state_lock.get(counterparty_node_id) {
 				Some(inner_state_lock) => {
 					let mut peer_state = inner_state_lock.lock().unwrap();
-					if let Some(jit_channel) = peer_state.outbound_channels_by_scid.get_mut(&scid) {
+					if let Some(lsps2) = peer_state.outbound_channels_by_scid.get_mut(&scid) {
```
johncantrell97: same as above, weird to lose channel as part of the name?

src/lsps2/channel_manager.rs
tnull: Ah, good point, seems my find/replace was a bit overeager and I didn't catch all the cases. Now reverted most of them in `channel_manager.rs` back to `jit_channel`.

src/lsps2/channel_manager.rs
tnull: Yeah, see above.

johncantrell97 (approved):

PR #51, opened 2023-11-28
expose compute_opening_fee util

johncantrell97:


src/lsps2/mod.rs
```diff
@@ -17,3 +17,4 @@ pub(crate) mod utils;
 
 pub use event::LSPS2Event;
 pub use msgs::{BuyResponse, GetInfoResponse, OpeningFeeParams, RawOpeningFeeParams};
+pub use utils::compute_opening_fee;
```
tnull, line 20: Rather than exposing this under the `lsps2` module, let's make `utils` `pub`.

tnull:
@johncantrell97 Any progress here? I think we can just make `lsps2/utils` `pub` instead of `pub(crate)` and be done with it?

tnull:
Closing as superseded by #83.

Issue #36, opened 2023-10-30
Project Restructuring: Further modularize codebase


tnull:
As discussed out of band it would be very useful if we could further modularize and expose the codebase at each of the individual levels. 

For example, we want to allow third parties to reuse our message types/serialization, or, just a specific LSPS implementation, without the need to use the full `LiquidityManager` and even without relying on LDK-specific `*Manager` types. To this end we'll need to make sure that not too many LDK types (in particular `ChannelManager`/`PeerManager` leak into the 'inner' part of the LSPS implementations.

PR #52, opened 2023-11-28
expose message type id

johncantrell97:
this became hidden in the refactor. it is needed when a user is writing their custom message handler.


johncantrell97:
hm I guess maybe user might not _need_ it. could just give _all_ custom messages to the handler and it will return Ok(None) if it doesn't know about the message type?

tnull:
> this became hidden in the refactor. it is needed when a user is writing their custom message handler.

Yeah, this was kind of intentional. We'll revisit where we should expose which types when tackling #36. We may want to expose said id then under `lsps0`.

tnull:
Now superseded by #58.

PR #56, opened 2023-12-05
Add initial `proptest` to check forwarding amount calculation

### LINKED ISSUE #45: LSPS2: Use `proptest` to test calculation utils
We want to explore using `proptest` to give us better test coverage of the many parameters the LSP (Client) could configure. 

Good first candidates to benefit from it would be our testing of `calculate_amount_to_forward` and `compute_opening_fee`.

tnull:
~~Based on lightningdevkit/lightning-liquidity#53~~, fixes lightningdevkit/lightning-liquidity#45.

We introduce an initial property-based test to check `calculate_amount_to_forward_per_htlc`. We also adapt the latter to be less susceptible to over/underflows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Pin crates for MSRV
         if: matrix.msrv
         run: |
-          # No need to pin currently
+          cargo update -p proptest --precise "1.2.0" --verbose # proptest 1.3.0 requires rustc 1.64.0
       - name: Cargo check
         run: cargo check --release
       - name: Check documentation
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,3 +22,6 @@ core2 = { version = "0.3.0", optional = true, default-features = false }
 chrono = { version = "0.4", default-features = false, features = ["serde", "alloc"] }
 serde = { version = "1.0", default-features = false, features = ["derive", "alloc"] }
 serde_json = "1.0"
+
+[dev-dependencies]
+proptest = "1.0.0"
--- a/proptest-regressions/lsps2/channel_manager.txt
+++ b/proptest-regressions/lsps2/channel_manager.txt
@@ -1,0 +1,6 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
--- a/src/lsps2/service.rs
+++ b/src/lsps2/service.rs
@@ -779,31 +779,38 @@ where
 fn calculate_amount_to_forward_per_htlc(
 	htlcs: &[InterceptedHTLC], total_amt_to_forward_msat: u64,
 ) -> Vec<(InterceptId, u64)> {
+	// TODO: we should eventually make sure the HTLCs are all above ChannelDetails::next_outbound_minimum_msat
 	let total_received_msat: u64 =
 		htlcs.iter().map(|htlc| htlc.expected_outbound_amount_msat).sum();
 
-	let mut fee_remaining_msat = total_received_msat - total_amt_to_forward_msat;
-	let total_fee_msat = fee_remaining_msat;
+	match total_received_msat.checked_sub(total_amt_to_forward_msat) {
+		Some(total_fee_msat) => {
+			let mut fee_remaining_msat = total_fee_msat;
 
-	let mut per_htlc_forwards = vec![];
+			let mut per_htlc_forwards = vec![];
 
-	for (index, htlc) in htlcs.iter().enumerate() {
-		let proportional_fee_amt_msat =
-			total_fee_msat * htlc.expected_outbound_amount_msat / total_received_msat;
+			for (index, htlc) in htlcs.iter().enumerate() {
+				let proportional_fee_amt_msat =
+					total_fee_msat * (htlc.expected_outbound_amount_msat / total_received_msat);
 
-		let mut actual_fee_amt_msat = core::cmp::min(fee_remaining_msat, proportional_fee_amt_msat);
-		fee_remaining_msat -= actual_fee_amt_msat;
+				let mut actual_fee_amt_msat =
+					core::cmp::min(fee_remaining_msat, proportional_fee_amt_msat);
+				fee_remaining_msat -= actual_fee_amt_msat;
 
-		if index == htlcs.len() - 1 {
-			actual_fee_amt_msat += fee_remaining_msat;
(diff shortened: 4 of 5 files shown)


src/lsps2/channel_manager.rs
```diff
@@ -1302,36 +1302,60 @@ fn calculate_amount_to_forward_per_htlc(
...
-				expected_outbound_amount_msat: 2000,
-			},
-			InterceptedHTLC {
-				intercept_id: InterceptId([2; 32]),
-				expected_outbound_amount_msat: 3000,
-			},
-		];
-
-		let total_amt_to_forward_msat = 5000;
+	const MAX_VALUE_MSAT: u64 = 21_000_000_0000_0000_000;
```
johncantrell97: odd formatting? could be 2_100_....?

src/lsps2/channel_manager.rs
```diff
@@ -1302,36 +1302,60 @@ fn calculate_amount_to_forward_per_htlc(
...
+			if total_received_msat < total_amt_to_forward_msat {
+				assert_eq!(result.len(), 0);
+			} else {
+				assert_ne!(result.len(), 0);
+				assert_eq!(result[0].0, htlcs[0].intercept_id);
+				assert_eq!(result[1].0, htlcs[1].intercept_id);
+				assert_eq!(result[2].0, htlcs[2].intercept_id);
+				assert!(result[0].1 <= o_0);
+				assert!(result[1].1 <= o_1);
+				assert!(result[2].1 <= o_2);
```
johncantrell97: probably should also assert that it's > 0. I think this is an edge case the current implementation will fail.

src/lsps2/channel_manager.rs
```diff
@@ -1302,36 +1302,60 @@ fn calculate_amount_to_forward_per_htlc(
...
+	const MAX_VALUE_MSAT: u64 = 21_000_000_0000_0000_000;
 
-		let result = calculate_amount_to_forward_per_htlc(&htlcs, total_amt_to_forward_msat);
-
-		assert_eq!(result[0].0, htlcs[0].intercept_id);
-		assert_eq!(result[0].1, 834);
+	fn arb_forward_amounts() -> impl Strategy<Value = (u64, u64, u64, u64)> {
+		(1u64..MAX_VALUE_MSAT, 1u64..MAX_VALUE_MSAT, 1u64..MAX_VALUE_MSAT, 1u64..MAX_VALUE_MSAT)
+			.prop_map(|(a, b, c, d)| {
+				(a, b, c, core::cmp::min(d, a.saturating_add(b).saturating_add(c)))
```
johncantrell97: cases where it d < a+b+c will all be the case where it can't extract a fee because received == amt_to_forward. i guess that's fine?

src/lsps2/channel_manager.rs
tnull: This is the usual formatting LDK uses internally. Pretty easy to parse visually (21_000_000 * 1_0000_0000 sats).

src/lsps2/channel_manager.rs
```diff
@@ -1302,36 +1302,60 @@ fn calculate_amount_to_forward_per_htlc(
...
+
+				let result_sum = result.iter().map(|(_, f)| f).sum::<u64>();
+				assert!(result_sum >= total_amt_to_forward_msat);
+				let five_pct = result_sum as f32 * 0.1;
+				let fair_share_0 = ((o_0 as f32 / total_received_msat as f32) * result_sum as f32).max(o_0 as f32);
+				assert!(result[0].1 as f32 <= fair_share_0 + five_pct);
+				let fair_share_1 = ((o_1 as f32 / total_received_msat as f32) * result_sum as f32).max(o_1 as f32);
+				assert!(result[1].1 as f32 <= fair_share_1 + five_pct);
+				let fair_share_2 = ((o_2 as f32 / total_received_msat as f32) * result_sum as f32).max(o_2 as f32);
+				assert!(result[2].1 as f32 <= fair_share_2 + five_pct);
```
johncantrell97: what happened to you not liking the casting to floats

src/lsps2/channel_manager.rs
tnull: In this case this is asserted as we generate arguments in `1..MAX_VALUE_MSAT `. Should I still add the assertion though?

src/lsps2/channel_manager.rs
tnull: Nothing, but in tests we can use floats to check our (much more efficient) integer based calculations match our prior expectations.

src/lsps2/channel_manager.rs
johncantrell97: yeah it's fine, just hadn't seen it like that before

src/lsps2/channel_manager.rs
johncantrell97: mm those are the htlc received amounts, you are asserting the amount to forward returned by the function being tested. it could return 0

tnull: > mm those are the htlc received amounts, you are asserting the amount to forward returned by the function being tested. it could return 0

Ah, you mean the forwarding amounts. Yes, it currently would return 0 for some inputs so we can't assert that. I don't  think we want to assert it either, but should eventually assert it's greater than the `htlc_minimum_msat`/`next_outbound_htlc_minimum`, as noted in the TODO left behind. I think this is out-of-scope for this PR though.

src/lsps2/channel_manager.rs
tnull: I think so? I think the LSP checks elsewhere that they'd get sufficient fees. In any case it's not really in scope for this testing PR.

src/lsps2/channel_manager.rs
tnull: Now also tracking here: https://github.com/lightningdevkit/lightning-liquidity/issues/59

tnull:
Rebased now that lightningdevkit/lightning-liquidity#53 landed and either addressed feedback or tracking it elsewhere.

src/lsps2/channel_manager.rs
johncantrell97: hm but returning 0 should be a bug since you can't forward a 0 value htlc? no? this is one reason I think this method will need to return a Result to actually handle all the edge cases properly

tnull: Right, but you also can't forward a 1-msat HTLC if your HTLC min is above that. We really have to check what we can forward per-HTLC and lower-bound accordingly, i.e., #59.

tnull:
Rebased after lightningdevkit/lightning-liquidity#58 landed. 

@johncantrell97 Let me know if I can merge.

tnull:
Merging as ACKed elsewhere

PR #71, opened 2023-12-15
allow user selected user_channel_id

### LINKED ISSUE #70: LSPS2: Let user provide a user_channel_id in invoice_parameters_generated
Currently this crate surfaces a user_channel_id to the user in OpenChannel event that they must use as the user_channel_id when they create the channel.  Currently, we are passing the SCID as the user_channel_id. This is because there's no way for the crate to map the channel during ChannelReady event to the scid used.

Not only is it a bit strange for our crate to specify what an ldk user's user_channel_id must be but it makes it difficult for them to track a entire flow from buy request to payment forwarded.

I think a better API would be one where the end-user can provide a user_channel_id to `invoice_parameters_generated` and our crate will resurface that during OpenChannel and keep a mapping from the user_channel_id => scid so that when ChannelReady event comes it can figure out the scid that was used.

johncantrell97:
Fixes #70

### DIFF
--- a/src/lsps2/event.rs
+++ b/src/lsps2/event.rs
@@ -126,7 +126,9 @@ pub enum LSPS2ServiceEvent {
 		amt_to_forward_msat: u64,
 		/// The fee earned for opening the channel.
 		opening_fee_msat: u64,
-		/// An internal id used to track channel open.
+		/// A user specified id used to track channel open.
 		user_channel_id: u128,
+		/// The short channel id to use in the route hint.
+		scid: u64,
 	},
 }
--- a/src/lsps2/event.rs
+++ b/src/lsps2/event.rs
@@ -52,8 +52,8 @@ pub enum LSPS2ClientEvent {
 	InvoiceGenerationReady {
 		/// The node id of the LSP.
 		counterparty_node_id: PublicKey,
-		/// The short channel id to use in the route hint.
-		scid: u64,
+		/// The intercept short channel id to use in the route hint.
+		intercept_scid: u64,
 		/// The `cltv_expiry_delta` to use in the route hint.
 		cltv_expiry_delta: u32,
 		/// The initial payment size you specified.
@@ -98,7 +98,7 @@ pub enum LSPS2ServiceEvent {
 	/// If `payment_size_msat` is [`Option::Some`] then the payer is allowed to use MPP.
 	/// If `payment_size_msat` is [`Option::None`] then the payer cannot use MPP.
 	///
-	/// You must generate an scid and `cltv_expiry_delta` for them to use
+	/// You must generate an intercept scid and `cltv_expiry_delta` for them to use
 	/// and call [`LSPS2ServiceHandler::invoice_parameters_generated`].
 	///
 	/// [`LSPS2ServiceHandler::invoice_parameters_generated`]: crate::lsps2::service::LSPS2ServiceHandler::invoice_parameters_generated
@@ -128,7 +128,7 @@ pub enum LSPS2ServiceEvent {
 		opening_fee_msat: u64,
 		/// A user specified id used to track channel open.
 		user_channel_id: u128,
-		/// The short channel id to use in the route hint.
-		scid: u64,
+		/// The intercept short channel id to use in the route hint.
+		intercept_scid: u64,
 	},
 }
--- a/src/lsps2/service.rs
+++ b/src/lsps2/service.rs
@@ -176,7 +176,7 @@ impl OutboundJITChannelState {
 
 struct OutboundJITChannel {
 	state: OutboundJITChannelState,
-	scid: u64,
+	intercept_scid: u64,
 	cltv_expiry_delta: u32,
 	client_trusts_lsp: bool,
 	user_channel_id: u128,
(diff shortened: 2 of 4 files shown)


src/lsps2/service.rs
```diff
@@ -240,14 +241,16 @@ impl OutboundJITChannel {
...
+	scid_by_user_channel_id: HashMap<u128, u64>,
 	pending_requests: HashMap<RequestId, LSPS2Request>,
 }
 
 impl PeerState {
 	fn new() -> Self {
 		let outbound_channels_by_scid = HashMap::new();
 		let pending_requests = HashMap::new();
-		Self { outbound_channels_by_scid, pending_requests }
+		let scid_by_user_channel_id = HashMap::new();
```
tnull: Not sure if we really need this additional `HashMap`? Couldn't we just store the `user_channel_id` in the `OutboundJitChannel`?

johncantrell97: We can't because with ChannelReady we only get the user channel id to lookup the OutboundJITChannel. 

Before this change we were making the user set user channel id to the scid but that's also strange and changed in this PR.

src/lsps2/event.rs
```diff
@@ -126,7 +126,9 @@ pub enum LSPS2ServiceEvent {
 		amt_to_forward_msat: u64,
 		/// The fee earned for opening the channel.
 		opening_fee_msat: u64,
-		/// An internal id used to track channel open.
+		/// A user specified id used to track channel open.
 		user_channel_id: u128,
+		/// The short channel id to use in the route hint.
+		scid: u64,
```
tnull: Can we start calling this `intercept_scid` here, and elsewhere for consistency with LDK? Would be much clearer that we're *never* talking about the 'real' scid here, but the fake SCIDs generated by LDK for us. This also needs to be reflected in the comment above to avoid misuse.

tnull (commented):
Hm, I'm generally wondering if we could switch to index everything by `user_channel_id` or `intercept_id` rather than `scid`. It seems the latter is really only ever necessary when constructing the `BuyResponse`, and using SCIDs in the interface is always confusing as it's unclear which one (real, inbound/outbound alias, intercept-scid) exactly is meant. I'm probably overlooking something here? Could we migrate away from tracking by SCID and just generate one when needed?

johncantrell97:
> Hm, I'm generally wondering if we could switch to index everything by `user_channel_id` or `intercept_id` rather than `scid`. It seems the latter is really only ever necessary when constructing the `BuyResponse`, and using SCIDs in the interface is always confusing as it's unclear which one (real, inbound/outbound alias, intercept-scid) exactly is meant. I'm probably overlooking something here? Could we migrate away from tracking by SCID and just generate one when needed? 

The issue is htlc intercepted only gives us scid and so it's the only way we can lookup what channel the htlc is for

tnull:
> > Hm, I'm generally wondering if we could switch to index everything by `user_channel_id` or `intercept_id` rather than `scid`. It seems the latter is really only ever necessary when constructing the `BuyResponse`, and using SCIDs in the interface is always confusing as it's unclear which one (real, inbound/outbound alias, intercept-scid) exactly is meant. I'm probably overlooking something here? Could we migrate away from tracking by SCID and just generate one when needed?
> 
> The issue is htlc intercepted only gives us scid and so it's the only way we can lookup what channel the htlc is for

Mhh, it would also provide the `intercept_id`, but I see that that is probably not an option. Could we then at least do `s/scid/intercept_scid/g` or so in the whole LSPS2 project (and adjust comments) to make it really clear *what* scid the user needs to plug in here?

johncantrell97:
> > > Hm, I'm generally wondering if we could switch to index everything by `user_channel_id` or `intercept_id` rather than `scid`. It seems the latter is really only ever necessary when constructing the `BuyResponse`, and using SCIDs in the interface is always confusing as it's unclear which one (real, inbound/outbound alias, intercept-scid) exactly is meant. I'm probably overlooking something here? Could we migrate away from tracking by SCID and just generate one when needed?
> > 
> > 
> > The issue is htlc intercepted only gives us scid and so it's the only way we can lookup what channel the htlc is for
> 
> Mhh, it would also provide the `intercept_id`, but I see that that is probably not an option. Could we then at least do `s/scid/intercept_scid/g` or so in the whole LSPS2 project (and adjust comments) to make it really clear _what_ scid the user needs to plug in here?

Yeah I think calling it intercept_scid makes sense since that is what the method is called on `ChannelManager` that they need to use to generate it.

johncantrell97:
@tnull ok, renamed scid to intercept_scid

tnull:
> @tnull ok, renamed scid to intercept_scid

Thanks! It seems there are two instances of `short_channel_id` left that we might want to align while we're at it (?): in `PendingPayment` and `invoice_params_received`

johncantrell97:
> > @tnull ok, renamed scid to intercept_scid
> 
> Thanks! It seems there are two instances of `short_channel_id` left that we might want to align while we're at it (?): in `PendingPayment` and `invoice_params_received`

Thanks, got them.

tnull (approved):

PR #73, opened 2023-12-20
Drop `APeerManager` dependency and replace with `process_msgs_callback`

### LINKED ISSUE #72: How to avoid cycle trait dependency in `LiquidityManager`?
It is nice when this [commit](https://github.com/lightningdevkit/lightning-liquidity/commit/ad3cc2f8c497225140c164040015494bbade3c76) comes in, the `LiquidityManager` now dependents on high-level traits and dependent on a lot of low-level things.  However, we ran into a problem after this.

The `LiquidityManager` now will need [`APeerManager`](https://docs.rs/lightning/0.0.118/lightning/ln/peer_handler/trait.APeerManager.html) as a dependency in trait. And [`PeerManager`](https://docs.rs/lightning/0.0.118/lightning/ln/peer_handler/struct.PeerManager.html#impl-APeerManager-for-PeerManager%3CDescriptor,+CM,+RM,+OM,+L,+CMH,+NS%3E) already implemented [`APeerManager`](https://docs.rs/lightning/0.0.118/lightning/ln/peer_handler/trait.APeerManager.html), but PeerManager also needs [`CustomMessageHandler`](https://docs.rs/lightning/0.0.118/lightning/ln/peer_handler/trait.CustomMessageHandler.html#).  When using the `LiquidityManager` as the [`CustomMessageHandler`](https://docs.rs/lightning/0.0.118/lightning/ln/peer_handler/trait.CustomMessageHandler.html#).  We will run into a cyclic trait dependency, and that is not allowed in Rust.

`PeerManager` <- `LiquidityManager` <- `CustomMessageHandler` <- `PeerManager`

Do you think there is a good suggestion to use this crate?  Many thanks for considering our use case.

tnull:
Fixes #72.

Previously, `LiquidityManager` (or at least its internal message queue) was holding a reference to `PeerManager`/`impl APeerManager` in order to be able to trigger message processing after new messages were enqueued.

However, this had two major drawbacks: firstly, it introduced an ugly (and hard to resolve) cycle in the type definitions as `PeerManager` depends on `CustomMessageHandler`, which in this case is `LiquidityManager`, itself depending on said `PeerManager` reference. Secondly, it introduced the complex `PeerManager` LDK object to the LSPS implementation which is otherwise not that depedent on LDK's internal types.

To resolve these issues, we heere replace the dependency on `PeerManager`/`impl APeerManager` with a simple generic callback that will be called every time new messages are enqueued.

Looking for feedback here. As a second step we could consider dropping the `MessageQueue` trait which was introduced as part of the prior refactor that moved the `APeerManager` bound from each individual message handler to the `DefaultMessageQueue`.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -23,7 +23,7 @@ use lightning::chain::{self, BestBlock, Confirm, Filter, Listen};
 use lightning::ln::channelmanager::{AChannelManager, ChainParameters};
 use lightning::ln::features::{InitFeatures, NodeFeatures};
 use lightning::ln::msgs::{ErrorAction, LightningError};
-use lightning::ln::peer_handler::{APeerManager, CustomMessageHandler};
+use lightning::ln::peer_handler::CustomMessageHandler;
 use lightning::ln::wire::CustomMessageReader;
 use lightning::sign::EntropySource;
 use lightning::util::logger::Level;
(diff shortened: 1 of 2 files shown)


src/manager.rs
```diff
@@ -208,47 +200,104 @@ where {
...
+	/// let process_msgs_callback = move || process_msgs_pm.process_events();
+	///
+	/// my_liquidity_manager.set_process_msgs_callback(process_msgs_callback);
+	/// # }
+	/// ```
+	///
+	/// [`PeerManager::process_events`]: lightning::ln::peer_handler::PeerManager::process_events
+	pub fn set_process_msgs_callback(&self, callback: impl Fn() + Send + Sync + 'static) {
+		self.pending_messages.set_process_msgs_callback(callback)
+	}
```
johncantrell97, line 299: I think the only reason we had this as something you set after construction was because of the circular reference issues with having it in the constructor? Could probably just move this into an optional callback param of the constructor?

johncantrell97 (approved):

src/manager.rs
tnull: I don't think so, as the circular reference is still there, it's just not reflected in the types themselves: before we can setup `PeerManager` we need to construct the `CustomMessageHandler`, whose callback (usually) needs to access a reference to said `PeerManager`.

PR #77, opened 2023-12-26
Documentation for lsps1

### LINKED ISSUE #67: LSPS1: Add missing docs
In #58 we added most of the docs missing from LSPS1.

However, we also left some `TODO`s, in particular in `lsps1/event.rs`. Moreover, we for now hid all `pub fn` from `LSPS1MessageHandler` for them to be exposed-and-documented as needed at a later date.

We should see to add the missing docs before removing the `cfg` guard.

Srg213:
Closes #67.

### DIFF
--- a/src/lsps1/client.rs
+++ b/src/lsps1/client.rs
@@ -237,13 +237,11 @@ where
 		}
 	}
 
-	/// Initiate the creation of an invoice that when paid will open a channel
-	/// with enough inbound liquidity to be able to receive the payment.
+	/// Retrieve information from the LSP regarding the options supported.
 	///
 	/// `counterparty_node_id` is the node_id of the LSP you would like to use.
 	///
-	/// `token` is an optional String that will be provided to the LSP.
-	/// It can be used by the LSP as an API key, coupon code, or some other way to identify a user.
+	/// 'channel_id' is the id used to uniquely identify the channel with counterparty node.
 	pub fn request_for_info(&self, counterparty_node_id: PublicKey, channel_id: u128) {
 		let channel = InboundCRChannel::new(channel_id);
 
@@ -321,9 +319,8 @@ where
 		Ok(())
 	}
 
-	/// Used by client to place an order to LSP with the provided parameters.
-	/// The client agrees to paying an channel fees as per requested by
-	/// the LSP.
+	/// Places an order with the connected LSP given its `counterparty_node_id`.
+	/// The client agrees to paying channel fees according to the provided parameters.
 	///
 	/// Should be called in response to receiving a [`LSPS1ClientEvent::GetInfoResponse`] event.
 	///
@@ -481,12 +478,12 @@ where
 		}
 	}
 
-	/// Used by client to check whether payment is received by LSP and status of order.
+	/// Queries the status of a pending payment, i.e., whether a payment has been received by the LSP.
 	///
 	/// Should be called in response to receiving a [`LSPS1ClientEvent::DisplayOrder`] event.
 	///
 	/// [`LSPS1ClientEvent::DisplayOrder`]: crate::lsps1::event::LSPS1ClientEvent::DisplayOrder
-	fn check_order_status(
+	pub fn check_order_status(
 		&self, counterparty_node_id: &PublicKey, order_id: OrderId, channel_id: u128,
 	) -> Result<(), APIError> {
 		let outer_state_lock = self.per_peer_state.write().unwrap();
--- a/src/lsps1/client.rs
+++ b/src/lsps1/client.rs
@@ -221,6 +221,7 @@ where
 	C::Target: Filter,
 	ES::Target: EntropySource,
 {
+	/// Constructs an `LSPS1ClientHandler`.
 	pub(crate) fn new(
 		entropy_source: ES, pending_messages: Arc<MessageQueue>, pending_events: Arc<EventQueue>,
 		channel_manager: CM, chain_source: Option<C>, config: LSPS1ClientConfig,
@@ -236,7 +237,14 @@ where
 		}
 	}
 
-	fn request_for_info(&self, counterparty_node_id: PublicKey, channel_id: u128) {
+	/// Initiate the creation of an invoice that when paid will open a channel
+	/// with enough inbound liquidity to be able to receive the payment.
+	///
+	/// `counterparty_node_id` is the node_id of the LSP you would like to use.
+	///
+	/// `token` is an optional String that will be provided to the LSP.
+	/// It can be used by the LSP as an API key, coupon code, or some other way to identify a user.
+	pub fn request_for_info(&self, counterparty_node_id: PublicKey, channel_id: u128) {
 		let channel = InboundCRChannel::new(channel_id);
 
 		let mut outer_state_lock = self.per_peer_state.write().unwrap();
@@ -313,7 +321,14 @@ where
 		Ok(())
 	}
 
-	fn place_order(
+	/// Used by client to place an order to LSP with the provided parameters.
+	/// The client agrees to paying an channel fees as per requested by
+	/// the LSP.
+	///
+	/// Should be called in response to receiving a [`LSPS1ClientEvent::GetInfoResponse`] event.
+	///
+	/// [`LSPS1ClientEvent::GetInfoResponse`]: crate::lsps1::event::LSPS1ClientEvent::GetInfoResponse
+	pub fn place_order(
 		&self, channel_id: u128, counterparty_node_id: &PublicKey, order: OrderParams,
 	) -> Result<(), APIError> {
 		let outer_state_lock = self.per_peer_state.write().unwrap();
@@ -466,6 +481,11 @@ where
 		}
 	}
 
+	/// Used by client to check whether payment is received by LSP and status of order.
+	///
+	/// Should be called in response to receiving a [`LSPS1ClientEvent::DisplayOrder`] event.
+	///
+	/// [`LSPS1ClientEvent::DisplayOrder`]: crate::lsps1::event::LSPS1ClientEvent::DisplayOrder
 	fn check_order_status(
 		&self, counterparty_node_id: &PublicKey, order_id: OrderId, channel_id: u128,
 	) -> Result<(), APIError> {
--- a/src/lsps1/event.rs
+++ b/src/lsps1/event.rs
@@ -1,3 +1,12 @@
+// This file is Copyright its original authors, visible in version control
+// history.
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
 //! Contains LSPS1 event types
 
 use super::msgs::{ChannelInfo, OptionsSupported, OrderId, OrderParams, OrderPayment};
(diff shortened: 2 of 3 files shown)


src/lsps1/client.rs
```diff
@@ -488,6 +503,11 @@ where
 		}
 	}
 
+	/// Used by client to check whether payment is received by LSP and status of order.
+	///
+	/// Should be called in response to receiving a [`LSPS1ClientEvent::DisplayOrder`] event.
+	///
+	/// [`LSPS1ClientEvent::DisplayOrder`]: crate::lsps1::event::LSPS1ClientEvent::DisplayOrder
 	fn check_order_status(
```
tnull: This needs to be `pub`.

src/lsps1/event.rs
```diff
@@ -10,64 +19,104 @@ use bitcoin::secp256k1::PublicKey;
 /// An event which an LSPS1 client should take some action in response to.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum LSPS1ClientEvent {
-	/// TODO
+	/// Information from the LSP about their supported protocol options.
+	///
+	/// You must check whether LSP supports the params the client wants and then call
+	/// [`LSPS1ClientHandler::place_order`] to place an order.
+	///
+	/// [`LSPS1ClientHandler::place_order`] : crate::lsps1::client::LSPS1ClientHandler::place_order
```
tnull: Remove superfluous whitespace:
```suggestion
	/// [`LSPS1ClientHandler::place_order`]: crate::lsps1::client::LSPS1ClientHandler::place_order
```

src/lsps1/event.rs
```diff
@@ -10,64 +19,104 @@ use bitcoin::secp256k1::PublicKey;
...
 	},
-	/// TODO
+	/// A request from client to check the status of the payment.
+	///
+	/// An event to poll for checking payment status either onchain or lightning.
+	///
+	/// You must call [`LSPS1ServiceHandler::update_order_status`] to update the client
+	/// regarding the status of the payment and order.
+	///
+	/// [`LSPS1ServiceHandler::send_invoice_for_order`]: crate::lsps1::service::LSPS1ServiceHandler::send_invoice_for_order
```
tnull: This needs to link to the right method:
```suggestion
	/// [`LSPS1ServiceHandler::update_order_status`]: crate::lsps1::service::LSPS1ServiceHandler::update_order_status
```

src/lsps1/client.rs
```diff
@@ -255,7 +256,14 @@ where
 		}
 	}
 
-	fn request_for_info(&self, counterparty_node_id: PublicKey, channel_id: u128) {
+	/// Initiate the creation of an invoice that when paid will open a channel
```
tnull: I think this is incorrect? Sending an `lsps1.info` request will just retrieve the LSP's supported options and parameters, not create an invoice, which only would happen in the `place_order` phase. Also unclear what "the payment" would refer to here and what `token` is meant. (I kinda suspect this was wrongly copied over from the LSPS2 docs?)

src/lsps1/client.rs
```diff
@@ -255,7 +256,14 @@ where
...
+	/// with enough inbound liquidity to be able to receive the payment.
+	///
+	/// `counterparty_node_id` is the node_id of the LSP you would like to use.
+	///
+	/// `token` is an optional String that will be provided to the LSP.
+	/// It can be used by the LSP as an API key, coupon code, or some other way to identify a user.
+	pub fn request_for_info(&self, counterparty_node_id: PublicKey, channel_id: u128) {
```
tnull, line 245: Btw, doesn't really need to happen here as this is a documentation-only PR, but I'd really like to renamed this as `request_for_info` isn't really a helpful API naming.

src/lsps1/client.rs
```diff
@@ -335,7 +343,14 @@ where
 		Ok(())
 	}
 
-	fn place_order(
+	/// Used by client to place an order to LSP with the provided parameters.
```
tnull:
```suggestion
	/// Places an order with the connected LSP given its `counterparty_node_id`.
```

src/lsps1/client.rs
```diff
@@ -335,7 +343,14 @@ where
...
+	/// The client agrees to paying an channel fees as per requested by
```
tnull:
```suggestion
	/// The client agrees to paying channel fees according to the provided parameters.
```

src/lsps1/client.rs
```diff
@@ -488,6 +503,11 @@ where
 		}
 	}
 
+	/// Used by client to check whether payment is received by LSP and status of order.
```
tnull:
```suggestion
	/// Queries the status of a pending payment, i.e., whether a payment has been received by the LSP.
```

src/lsps1/event.rs
```diff
@@ -10,64 +19,104 @@ use bitcoin::secp256k1::PublicKey;
 /// An event which an LSPS1 client should take some action in response to.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum LSPS1ClientEvent {
-	/// TODO
+	/// Information from the LSP about their supported protocol options.
+	///
+	/// You must check whether LSP supports the params the client wants and then call
```
tnull:
```suggestion
	/// You must check whether LSP supports the parameters the client wants and then call
```

src/lsps1/event.rs
```diff
@@ -10,64 +19,104 @@ use bitcoin::secp256k1::PublicKey;
...
+	/// [`LSPS1ClientHandler::place_order`] to place an order.
+	///
+	/// [`LSPS1ClientHandler::place_order`] : crate::lsps1::client::LSPS1ClientHandler::place_order
 	GetInfoResponse {
-		/// TODO
+		/// This is a randomly generated identifier used to track the channel state.
```
tnull, line 29: nit: Leave one line whitespace after initial description line.
```suggestion
		/// This is a randomly generated identifier used to track the channel state.
		///
```

src/lsps1/event.rs
```diff
@@ -10,64 +19,104 @@ use bitcoin::secp256k1::PublicKey;
...
+		/// It is not related in anyway to the eventual lightning channel id.
+		/// It needs to be passed to [`LSPS1ClientHandler::place_order`].
+		///
+		/// [`LSPS1ClientHandler::place_order`]: crate::lsps1::client::LSPS1ClientHandler::place_order
 		id: u128,
-		/// TODO
+		/// An identifier to track messages received.
```
tnull, line 36: Why do we need to expose this request id in the events? What do we expect the user to do with it?

src/lsps1/event.rs
```diff
@@ -10,64 +19,104 @@ use bitcoin::secp256k1::PublicKey;
...
+		/// The website of the LSP.
 		website: String,
-		/// TODO
+		/// All options supported by the LSP.
 		options_supported: OptionsSupported,
 	},
-	/// TODO
+	/// Confirmation from the LSP about the order created by the client.
+	///
+	/// When the invoice is paid, the LSP will open a channel to you
```
tnull: Note that we don't only support Lightning payments in LSPS1, so this should probably speak of payments more generally.

src/lsps1/event.rs
```diff
@@ -10,64 +19,104 @@ use bitcoin::secp256k1::PublicKey;
...
+		/// The details regarding state of the channel ordered.
 		channel: Option<ChannelInfo>,
 	},
 }
 
 /// An event which an LSPS1 server should take some action in response to.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum LSPS1ServiceEvent {
-	/// TODO
+	/// A client has selected the parameters to use from the supported options of LSP
```
tnull:
```suggestion
	/// A client has selected the parameters to use from the supported options of the LSP
```

src/lsps1/event.rs
```diff
@@ -10,64 +19,104 @@ use bitcoin::secp256k1::PublicKey;
...
+	/// and would like to open a channel with the given invoice.
+	///
+	/// You must call [`LSPS1ServiceHandler::send_invoice_for_order`] to
```
tnull, line 79: How does this work with the onchain payment flow?

src/lsps1/service.rs
```diff
@@ -254,7 +255,12 @@ where
 		Ok(())
 	}
 
-	fn send_invoice_for_order(
+	/// Used by LSP to send invoice containing details regarding the channel fees and payment information.
```
tnull, line 236: I'm confused: this doesn't actually send and invoice (which seems more like the LSPS2-flow anyways), but simply a `CreateOrderResponse`. What am I missing?

tnull (commented):
Thanks! Did a first pass, here are a few questions.

It also unfortunately seems that our CI doc checks were not working correctly with `cfg(lsps1)`, which is now fixed on `main` (see #78/#79). Could you rebase on `main` and make sure CI passes?

src/lsps1/client.rs
Srg213: Yes.

src/lsps1/client.rs
Srg213: Right. My bad.

src/lsps1/event.rs
Srg213: Unfortunately, I used the wrong naming convention at the start and then forgot to rename this.
I will rename some of the functions to remove ambiguities.

src/lsps1/service.rs
Srg213: This simply sends a` CreateOrderResponse`. Will rename the functions.

src/lsps1/client.rs
Srg213: Renamed to `send_get_info_request`

src/lsps1/event.rs
Srg213: Right. It should not be exposed to user. Removed.

src/lsps1/service.rs
Srg213: Renamed to `send_payment_details`.

src/lsps1/event.rs
Srg213: Renamed the event to `Request_for_payment_details`

src/lsps1/client.rs
tnull: Yeah, probably fine for now. When this is getting closer to be ready for use, we might want to consider simplifying the flow a bit and making the naming a bit more intuitive (i.e., choose names that don't require the user to be fully aware of the LSPS1 protocol).

src/lsps1/client.rs
```diff
@@ -300,7 +300,7 @@ where
 
 				self.pending_events.enqueue(Event::LSPS1Client(LSPS1ClientEvent::GetInfoResponse {
 					id: channel_id,
-					request_id,
+
```
tnull: nit: Delete newly introduced blank line.

src/lsps1/client.rs
```diff
@@ -294,7 +300,7 @@ where
 
 				self.pending_events.enqueue(Event::LSPS1Client(LSPS1ClientEvent::GetInfoResponse {
 					id: channel_id,
```
tnull: These `id`/`channel_id` fields should be renamed `user_channel_id` for consistency reasons. However, let's not do this in this doc PR, now tracking here: https://github.com/lightningdevkit/lightning-liquidity/issues/89

tnull (commented):
Generally LGTM, mod one nit and the CI failing.

Srg213:
Added commits only related to docs. I will create a new PR for api fixes.
Please approve this PR.

src/lsps1/client.rs
```diff
@@ -236,7 +237,12 @@ where
 		}
 	}
 
-	fn request_for_info(&self, counterparty_node_id: PublicKey, channel_id: u128) {
+	/// Retrieve information from the LSP regarding the options supported.
+	///
+	/// `counterparty_node_id` is the node_id of the LSP you would like to use.
+	///
+	/// 'channel_id' is the id used to uniquely identify the channel with counterparty node.
```
tnull, line 244: nit: Wrong ticks used here:
```suggestion
	/// `channel_id` is the id used to uniquely identify the channel with counterparty node.
```

But happy to fix this in a follow up sometime.

tnull (approved):

PR #74, opened 2023-12-20
Upgrade LDK to 0.0.119 and rust-bitcoin to 0.30.2

tnull:

PR #80, opened 2024-01-03
Implement `next_event_async` method polling the event queue

tnull:
Based on #74.

We implement a way to asynchronously poll the queue for new events,
providing an async alternative to `wait_next_event`.

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -128,3 +128,112 @@ impl Future for EventFuture {
 		}
 	}
 }
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use crate::lsps0::event::LSPS0ClientEvent;
+	use bitcoin::secp256k1::{PublicKey, Secp256k1, SecretKey};
+
+	#[tokio::test]
+	#[cfg(feature = "std")]
+	async fn event_queue_works() {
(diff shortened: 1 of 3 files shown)


src/events.rs
```diff
@@ -65,6 +77,10 @@ impl EventQueue {
 		drop(queue);
 
 		if should_notify {
+			if let Some(waker) = self.waker.lock().unwrap().take() {
+				waker.wake();
```
johncantrell97, line 81: this feels a little weird, can you help me understand? is this some weird edge-case where the user is using _both_ the sync `wait_next_event` and async in tandem somehow / for some reason?

johncantrell97 (approved):

src/events.rs
tnull: Yeah, kinda. Users likely shouldn't use both at the same time, but here we just make sure to wake the `Waker` to poll again if we still have events left to be processed, just as we wake up any parked threads via the `Condvar`. Tbh. I'm not entirely sure this is necessary, but also shouldn't hurt, IIUC.

I now added a test for the event queue operations that has at least some checks for mixing matching blocking/async APIs.

tnull:
Added a test case for the event queue operations. I'll go ahead and merge this as this only confirms the previously-ACKed code.

PR #83, opened 2024-01-04
Expose `lsps2/utils.rs`

tnull:
Supersedes #51

We expose the LSPS2 utils .. as they are useful and already properly documented for the most part.


tnull:
I'm going ahead and merge this as it's a trivial change.

PR #88, opened 2024-01-08
LSPS2: Improve client-side flow API

tnull:
We improve the LSPS2 client-side API by making two changes:

1) We drop the `client_trusts_lsp` field from the event and rather allow the user to configure the flag via `LSPS2ClientConfig`.

2) In order to make the API flow a bit more intuitive, we rename the
corresponding methods and events:

- `create_invoice` -> `request_opening_params`
- `opening_fee_params_selected` -> `select_opening_params`
- `LSPS2ClientEvent::GetInfoResponse` -> `OpeningParametersReady`
- `LSPS2ClientEvent::InvoiceGenerationReady` -> `InvoiceParametersReady`

We also cleanup the docs a bit.

### DIFF
--- a/src/lsps2/client.rs
+++ b/src/lsps2/client.rs
@@ -13,7 +13,7 @@ use crate::events::{Event, EventQueue};
 use crate::lsps0::msgs::{ProtocolMessageHandler, RequestId, ResponseError};
 use crate::lsps2::event::LSPS2ClientEvent;
 use crate::message_queue::MessageQueue;
-use crate::prelude::{HashMap, String, ToString};
+use crate::prelude::{HashMap, HashSet, String};
 use crate::sync::{Arc, Mutex, RwLock};
 
 use lightning::ln::msgs::{ErrorAction, LightningError};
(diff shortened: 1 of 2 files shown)


tnull:
@johncantrell97 Please me know if the first point somehow conflicts with the LSP-trusts-client flow you had in mind.

src/lsps2/client.rs
```diff
@@ -405,6 +431,18 @@ where
...
+				// it.
+				if !self.config.client_trusts_lsp && result.client_trusts_lsp {
+					peer_state.remove_inbound_channel(jit_channel_id);
+					return Err(LightningError {
+						err: format!(
+							"Aborting JIT channel flow as the LSP requires 'client_trusts_lsp' mode, which we disallow"
+						),
+						action: ErrorAction::IgnoreAndLog(Level::Info),
+					});
+				}
```
johncantrell97, line 337: Hm, there's currently no way for the end-user client to detect this happened, right? Might need to surface some kind of error event or similar to the user so they can surface to wallet user (or retry with different lsp or however they want to handle it).

In mutiny's current implementation this would just end up with a infinite spinner.

johncantrell97: Fine with this being a follow up though

johncantrell97 (approved):
This looks good to me, like the renames. I don't see why it would conflict with any of the client_trusts_lsp stuff I did on server side.  Just means the flow fails if client requires it to be false.

src/lsps2/client.rs
tnull: Mhh, good point. I think given this is not the only possible failure case we may indeed need to introduce a `ChannelFailed` event

tnull: Now tracking here: https://github.com/lightningdevkit/lightning-liquidity/issues/90

tnull:
Now added another commit that simplifies the client-side tracking a lot. Let me know if it checks out.

johncantrell97 (approved):

tnull:
Just confirmed this still works as expected. Merging it now.

PR #91, opened 2024-01-10
Check variable-amount payments against `{min,max}_payment_size_msat`

tnull:
Previously, we would open a channel of arbitrary size if the user didn't supply the optional `payment_size_msat`, i.e., would use `no-MPP+var-invoice` mode.

Here, we make sure to check the resulting `expected_payment_size_msat` would always adhere to our set limits and fail the HTLC otherwise.


johncantrell97 (approved):
good catch, looks good to me!

PR #102, opened 2024-02-12
fix prepayment probing by sequencing subsequent payments

johncantrell97:
This replaces the `htlcs: Vec<InterceptedHTLC>` that is stored in the OutboundJITChannelState with a map that collects htlcs grouped by PaymentHash `payments: HashMap<PaymentHash, Vec<InterceptedHTLC>>.  It also introduces a queue of payments `payment_queue: VecDeque<PaymentHash>`.  We only store the PaymentHash in the queue and can lookup the htlcs that make up the payments in the map.

When we first transition from `AwaitingPayment` to `PendingChannelOpen` we create the queue and make sure the payment hash of the htlc that triggered the transition is in the front of the queue.

When we reach ChannelReady we can pop the first payment off the queue and proceed with our normal logic of forwarding all of the HTLCs that make up that payment.

At any point in the flow if extra HTLCs come in, they will get added to the map and queue in the order they arrived.

We introduce a new public function `htlc_handling_failed` where the user can pass through the `failed_next_destination` from `HTLCHandlingFailed` events.  When we receive an `HTLCDestination::InvalidForward` we can lookup the OutboundJITChannel via the `requested_forward_scid` property.

We pop the next payment off the queue and attempt to forward it the same way we forward during ChannelReady.

It feels like there's still some issues here around subsequent payments that are different sizes.  This assumes the initial payment and subsequent ones are all the same size and therefore the same opening fee is attempted to be extracted.  This will work in the prepayment probe case we are solving for but not the general case.

To solve the general case I think we will need to store and/or recalculate both the opening fee and total amount to be forwarded based on the collected htlcs for the next payment.  

I guess we also need to handle `PaymentForwarded` events and maintain a map from ChannelId to scid so that we can forward the rest of the payment queue once we have forwarded a payment successfully.

### DIFF
--- a/src/lsps2/payment_queue.rs
+++ b/src/lsps2/payment_queue.rs
@@ -1,0 +1,119 @@
+use crate::prelude::Vec;
+use lightning::ln::channelmanager::InterceptId;
+use lightning::ln::PaymentHash;
+
+/// Holds payments with the corresponding HTLCs until it is possible to pay the fee.
+/// When the fee is successfully paid with a forwarded payment, the queue should be consumed and the
+/// remaining payments forwarded.
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub(crate) struct PaymentQueue {
+	payments: Vec<(PaymentHash, Vec<InterceptedHTLC>)>,
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub(crate) struct InterceptedHTLC {
+	pub(crate) intercept_id: InterceptId,
+	pub(crate) expected_outbound_amount_msat: u64,
+	pub(crate) payment_hash: PaymentHash,
+}
+
+impl PaymentQueue {
+	pub(crate) fn new() -> PaymentQueue {
+		PaymentQueue { payments: Vec::new() }
+	}
+
+	pub(crate) fn add_htlc(&mut self, new_htlc: InterceptedHTLC) -> (u64, usize) {
+		let payment = self.payments.iter_mut().find(|(p, _)| p == &new_htlc.payment_hash);
+		if let Some((payment_hash, htlcs)) = payment {
+			// HTLCs within a payment should have the same payment hash.
+			debug_assert!(htlcs.iter().all(|htlc| htlc.payment_hash == *payment_hash));
+			// The given HTLC should not already be present.
+			debug_assert!(htlcs.iter().all(|htlc| htlc.intercept_id != new_htlc.intercept_id));
+			htlcs.push(new_htlc);
+			let total_expected_outbound_amount_msat =
+				htlcs.iter().map(|htlc| htlc.expected_outbound_amount_msat).sum();
+			(total_expected_outbound_amount_msat, htlcs.len())
+		} else {
+			let expected_outbound_amount_msat = new_htlc.expected_outbound_amount_msat;
+			self.payments.push((new_htlc.payment_hash, vec![new_htlc]));
+			(expected_outbound_amount_msat, 1)
+		}
+	}
+
+	pub(crate) fn pop_greater_than_msat(
+		&mut self, amount_msat: u64,
+	) -> Option<(PaymentHash, Vec<InterceptedHTLC>)> {
+		let position = self.payments.iter().position(|(_payment_hash, htlcs)| {
+			htlcs.iter().map(|htlc| htlc.expected_outbound_amount_msat).sum::<u64>() >= amount_msat
+		});
+		position.map(|position| self.payments.remove(position))
+	}
+
+	pub(crate) fn clear(&mut self) -> Vec<InterceptedHTLC> {
+		self.payments.drain(..).map(|(_k, v)| v).flatten().collect()
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn test_payment_queue() {
+		let mut payment_queue = PaymentQueue::new();
+		assert_eq!(
+			payment_queue.add_htlc(InterceptedHTLC {
+				intercept_id: InterceptId([0; 32]),
+				expected_outbound_amount_msat: 200_000_000,
+				payment_hash: PaymentHash([100; 32]),
+			}),
+			(200_000_000, 1),
+		);
+		assert_eq!(payment_queue.pop_greater_than_msat(500_000_000), None);
+
+		assert_eq!(
+			payment_queue.add_htlc(InterceptedHTLC {
+				intercept_id: InterceptId([1; 32]),
+				expected_outbound_amount_msat: 300_000_000,
+				payment_hash: PaymentHash([101; 32]),
+			}),
+			(300_000_000, 1),
+		);
+		assert_eq!(payment_queue.pop_greater_than_msat(500_000_000), None);
+
+		assert_eq!(
+			payment_queue.add_htlc(InterceptedHTLC {
+				intercept_id: InterceptId([2; 32]),
+				expected_outbound_amount_msat: 300_000_000,
+				payment_hash: PaymentHash([100; 32]),
+			}),
+			(500_000_000, 2),
+		);
+		assert_eq!(
+			payment_queue.pop_greater_than_msat(500_000_000),
+			Some((
+				PaymentHash([100; 32]),
+				vec![
+					InterceptedHTLC {
+						intercept_id: InterceptId([0; 32]),
+						expected_outbound_amount_msat: 200_000_000,
+						payment_hash: PaymentHash([100; 32]),
+					},
+					InterceptedHTLC {
+						intercept_id: InterceptId([2; 32]),
+						expected_outbound_amount_msat: 300_000_000,
+						payment_hash: PaymentHash([100; 32]),
+					},
+				]
+			))
+		);
+		assert_eq!(
+			payment_queue.clear(),
+			vec![InterceptedHTLC {
+				intercept_id: InterceptId([1; 32]),
+				expected_outbound_amount_msat: 300_000_000,
+				payment_hash: PaymentHash([101; 32]),
+			}]
+		);
+	}
+}
--- a/src/lsps2/service.rs
+++ b/src/lsps2/service.rs
@@ -12,6 +12,7 @@
 use crate::events::{Event, EventQueue};
 use crate::lsps0::ser::{ProtocolMessageHandler, RequestId, ResponseError};
 use crate::lsps2::event::LSPS2ServiceEvent;
+use crate::lsps2::payment_queue::{InterceptedHTLC, PaymentQueue};
 use crate::lsps2::utils::{compute_opening_fee, is_valid_opening_fee_params};
 use crate::message_queue::MessageQueue;
 use crate::prelude::{HashMap, String, ToString, Vec};
@@ -19,7 +20,7 @@ use crate::sync::{Arc, Mutex, RwLock};
 
 use lightning::ln::channelmanager::{AChannelManager, InterceptId};
 use lightning::ln::msgs::{ErrorAction, LightningError};
-use lightning::ln::ChannelId;
+use lightning::ln::{ChannelId, PaymentHash};
 use lightning::util::errors::APIError;
 use lightning::util::logger::Level;
 
@@ -45,10 +46,19 @@ pub struct LSPS2ServiceConfig {
 	pub promise_secret: [u8; 32],
 }
 
-#[derive(Copy, Clone, PartialEq, Eq, Debug)]
-struct InterceptedHTLC {
-	intercept_id: InterceptId,
-	expected_outbound_amount_msat: u64,
+/// Information about the initial payment size and JIT channel opening fee.
+/// This will be provided in the `OpenChannel` event.
+#[derive(Clone, Debug, PartialEq)]
+struct OpenChannelParams {
+	opening_fee_msat: u64,
+	amt_to_forward_msat: u64,
+}
+
+/// A payment that will be forwarded while skimming the given JIT channel opening fee.
+#[derive(Clone, Debug, PartialEq)]
+struct FeePayment {
+	htlcs: Vec<InterceptedHTLC>,
+	opening_fee_msat: u64,
 }
 
 struct ChannelStateError(String);
@@ -59,61 +69,44 @@ impl From<ChannelStateError> for LightningError {
 	}
 }
 
-#[derive(PartialEq, Debug)]
+/// The different states a requested JIT channel can be in.
+#[derive(Debug)]
 enum OutboundJITChannelState {
-	AwaitingPayment {
-		min_fee_msat: u64,
-		proportional_fee: u32,
-		min_payment_size_msat: u64,
-		max_payment_size_msat: u64,
-		htlcs: Vec<InterceptedHTLC>,
-		payment_size_msat: Option<u64>,
-	},
-	PendingChannelOpen {
-		htlcs: Vec<InterceptedHTLC>,
-		opening_fee_msat: u64,
-		amt_to_forward_msat: u64,
-	},
-	ChannelReady {
-		htlcs: Vec<InterceptedHTLC>,
-		opening_fee_msat: u64,
-	},
+	/// The JIT channel SCID was created after a buy request, and we are awaiting an initial payment
+	/// of sufficient size to open the channel.
+	PendingInitialPayment { payment_queue: Arc<Mutex<PaymentQueue>> },
+	/// An initial payment of sufficient size was intercepted to the JIT channel SCID, triggering the
+	/// opening of the channel. We are awaiting the completion of the channel establishment.
+	PendingChannelOpen { payment_queue: Arc<Mutex<PaymentQueue>>, opening_fee_msat: u64 },
+	/// The channel is open and a payment was forwarded while skimming the JIT channel fee.
+	PendingPaymentForward { payment_queue: Arc<Mutex<PaymentQueue>>, _opening_fee_msat: u64 },
+	/// The channel is open and a payment was successfully forwarded while skimming the JIT channel
+	/// fee. Any subsequent HTLCs can be forwarded without additional logic.
+	PaymentForwarded,
 }
 
 impl OutboundJITChannelState {
-	fn new(payment_size_msat: Option<u64>, opening_fee_params: OpeningFeeParams) -> Self {
-		OutboundJITChannelState::AwaitingPayment {
-			min_fee_msat: opening_fee_params.min_fee_msat,
-			proportional_fee: opening_fee_params.proportional,
-			min_payment_size_msat: opening_fee_params.min_payment_size_msat,
-			max_payment_size_msat: opening_fee_params.max_payment_size_msat,
-			htlcs: vec![],
-			payment_size_msat,
+	fn new() -> Self {
+		OutboundJITChannelState::PendingInitialPayment {
+			payment_queue: Arc::new(Mutex::new(PaymentQueue::new())),
 		}
 	}
 
-	fn htlc_intercepted(&self, htlc: InterceptedHTLC) -> Result<Self, ChannelStateError> {
+	fn htlc_intercepted(
+		&mut self, opening_fee_params: &OpeningFeeParams, payment_size_msat: &Option<u64>,
+		htlc: InterceptedHTLC,
+	) -> Result<(Self, Option<OpenChannelParams>), ChannelStateError> {
 		match self {
-			OutboundJITChannelState::AwaitingPayment {
-				htlcs,
-				payment_size_msat,
-				min_payment_size_msat,
-				max_payment_size_msat,
-				min_fee_msat,
-				proportional_fee,
-			} => {
-				let mut htlcs = htlcs.clone();
-				htlcs.push(htlc);
-
-				let total_expected_outbound_amount_msat =
-					htlcs.iter().map(|htlc| htlc.expected_outbound_amount_msat).sum();
+			OutboundJITChannelState::PendingInitialPayment { payment_queue } => {
+				let (total_expected_outbound_amount_msat, num_htlcs) =
+					payment_queue.lock().unwrap().add_htlc(htlc);
 
 				let (expected_payment_size_msat, mpp_mode) =
 					if let Some(payment_size_msat) = payment_size_msat {
 						(*payment_size_msat, true)
 					} else {
-						debug_assert_eq!(htlcs.len(), 1);
-						if htlcs.len() != 1 {
+						debug_assert_eq!(num_htlcs, 1);
+						if num_htlcs != 1 {
 							return Err(ChannelStateError(
 								format!("Paying via multiple HTLCs is disallowed in \"no-MPP+var-invoice\" mode.")
 							));
@@ -121,25 +114,25 @@ impl OutboundJITChannelState {
 						(total_expected_outbound_amount_msat, false)
 					};
 
-				if expected_payment_size_msat < *min_payment_size_msat
-					|| expected_payment_size_msat > *max_payment_size_msat
+				if expected_payment_size_msat < opening_fee_params.min_payment_size_msat
+					|| expected_payment_size_msat > opening_fee_params.max_payment_size_msat
 				{
 					return Err(ChannelStateError(
 							format!("Payment size violates our limits: expected_payment_size_msat = {}, min_payment_size_msat = {}, max_payment_size_msat = {}",
 									expected_payment_size_msat,
-									min_payment_size_msat,
-									max_payment_size_msat
+									opening_fee_params.min_payment_size_msat,
+									opening_fee_params.max_payment_size_msat
 							)));
 				}
 
 				let opening_fee_msat = compute_opening_fee(
 					expected_payment_size_msat,
-					*min_fee_msat,
-					(*proportional_fee).into(),
+					opening_fee_params.min_fee_msat,
+					opening_fee_params.proportional.into(),
 				).ok_or(ChannelStateError(
 					format!("Could not compute valid opening fee with min_fee_msat = {}, proportional = {}, and expected_payment_size_msat = {}",
-						min_fee_msat,
-						proportional_fee,
+						opening_fee_params.min_fee_msat,
+						opening_fee_params.proportional,
 						expected_payment_size_msat
 					)
 				))?;
@@ -151,21 +144,20 @@ impl OutboundJITChannelState {
 				if total_expected_outbound_amount_msat >= expected_payment_size_msat
 					&& amt_to_forward_msat > 0
 				{
-					Ok(OutboundJITChannelState::PendingChannelOpen {
-						htlcs,
+					let pending_channel_open = OutboundJITChannelState::PendingChannelOpen {
+						payment_queue: Arc::clone(&payment_queue),
 						opening_fee_msat,
-						amt_to_forward_msat,
-					})
+					};
+					let open_channel_params =
+						OpenChannelParams { opening_fee_msat, amt_to_forward_msat };
+					Ok((pending_channel_open, Some(open_channel_params)))
 				} else {
 					if mpp_mode {
-						Ok(OutboundJITChannelState::AwaitingPayment {
-							min_fee_msat: *min_fee_msat,
-							proportional_fee: *proportional_fee,
-							htlcs,
-							payment_size_msat: *payment_size_msat,
-							min_payment_size_msat: *min_payment_size_msat,
-							max_payment_size_msat: *max_payment_size_msat,
-						})
+						let pending_initial_payment =
+							OutboundJITChannelState::PendingInitialPayment {
+								payment_queue: Arc::clone(&payment_queue),
+							};
+						Ok((pending_initial_payment, None))
 					} else {
 						Err(ChannelStateError(
 							"Intercepted HTLC is too small to pay opening fee".to_string(),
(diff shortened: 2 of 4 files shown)


src/lsps2/service.rs
```diff
@@ -69,17 +72,21 @@ enum OutboundJITChannelState {
 	AwaitingPayment {
 		min_fee_msat: u64,
 		proportional_fee: u32,
-		htlcs: Vec<InterceptedHTLC>,
 		payment_size_msat: Option<u64>,
+		payments: HashMap<PaymentHash, Vec<InterceptedHTLC>>,
```
tnull: As discussed elsewhere, this should already be a queue here. Also, let's not keep the `payments` and `payment_queue` separate, but just implement a `PaymentQueue` in a new `queue.rs` that offers the expected behavior (i.e., FIFO queue grouped by payment hash) and we'd hold an `Arc<Mutex<PaymentQueue>>` across states. It will also make it much easier to maintain a consistent view, especially when we come to persistence.

src/lsps2/service.rs
```diff
@@ -69,17 +72,21 @@ enum OutboundJITChannelState {
...
+		payment_queue: VecDeque<PaymentHash>,
 		opening_fee_msat: u64,
 		amt_to_forward_msat: u64,
 	},
 	ChannelReady {
-		htlcs: Vec<InterceptedHTLC>,
+		payments: HashMap<PaymentHash, Vec<InterceptedHTLC>>,
+		payment_queue: VecDeque<PaymentHash>,
+		opening_fee_msat: u64,
```
tnull, line 115: The additions of `opening_fee_msat` and `channel_id` seem like (more or less) unrelated changes, please pull them out into their own commit at the very least. Given that this PR will probably will grow a bit, you might want to introduce  them in a separate PR altogether.

src/lsps2/service.rs
```diff
@@ -509,6 +605,70 @@ where
...
 	}
 
+	/// Forward [`Event::HTLCHandlingFailed`] event parameter into this function.
+	///
+	/// Will attempt to forward the next payment in the queue if one is present.
+	/// Will do nothing if the intercept scid does not match any of the ones we gave out
+	/// or if the payment queue is empty
+	///
+	/// [`Event::HTLCHandlingFailed`]: lightning::events::Event::HTLCHandlingFailed
+	pub fn htlc_handling_failed(
```
tnull, line 728: Likewise, the failure case should be added in a separate commit at the very least.

johncantrell97:
Note: Just realized there's a timing issue if we are in ChannelReady state and an HTLC fails (or succeeds for that matter) before the next payment/htlc arrives then there is nothing to trigger the subsequent forward.

We probably need some concept of or flag for "is_pending_payment" and immediately forward if there is no currently pending payment.

tnull:
> Note: Just realized there's a timing issue if we are in ChannelReady state and an HTLC fails (or succeeds for that matter) before the next payment/htlc arrives then there is nothing to trigger the subsequent forward.
> 
> We probably need some concept of or flag for "is_pending_payment" and immediately forward if there is no currently pending payment.

Not sure I understand the issue exactly. IIUC we should be fine if we default to forwarding the HTLC in `htlc_intercepted`, except when we're in `AwaitingPayment` state, in which case we'd follow the current "initial forward" path, collecting MPPs, etc?

src/lsps2/service.rs
wvanlint: Pulled out HTLC queueing logic into `payment_queue.rs`.

src/lsps2/service.rs
wvanlint: Separated these changes from the `PaymentQueue` refactoring.

wvanlint:
Will be picking this up while John is away, I added a revert to maintain history and will squash/rebase the changes at a later point.
I separated the `PaymentQueue` refactoring as a first commit maintaining the original behavior, and will add handling around `htlc_handling_failed` and `payment_forwarded` in following commits.

> Not sure I understand the issue exactly. IIUC we should be fine if we default to forwarding the HTLC in htlc_intercepted, except when we're in AwaitingPayment state, in which case we'd follow the current "initial forward" path, collecting MPPs, etc?

I believe the issue is that we need to collect the JIT channel fee exactly one time with a successful payment. If the first collected payment fails (e.g. due to being a probe), we will need to deduct fees from a second payment which might arrive later in parts.

tnull:
> Will be picking this up while John is away, I added a revert to maintain history and will squash/rebase the changes at a later point. I separated the `PaymentQueue` refactoring as a first commit maintaining the original behavior, and will add handling around `htlc_handling_failed` and `payment_forwarded` in following commits.

Alright, sounds good. Will probably only get around next week to take a closer look/do a full pass. However, in the meantime this will need a rebase after #97 landed, i.e., `{min,max}_payment_size_msat` are now part of `OpeningFeeParams` and hence are per-payment limits. FWIW, we could just hold the full opening fee params rather than passing the individual fields around.

> I believe the issue is that we need to collect the JIT channel fee exactly one time with a successful payment. If the first collected payment fails (e.g. due to being a probe), we will need to deduct fees from a second payment which might arrive later in parts.

Yes, that's the general issue here, I'm however still not fully getting why @johncantrell97 suggested we need a `is_pending_payment` flag and where/how it would be used.

johncantrell97:
I think the issue I'm getting at comes down to how the sequencer works. The idea is that there are only the following triggers that will attempt to forward the next payment from the queue:

1) transition to ChannelReady state when we first open the channel. 

2) htlc handling failed, when a forward fails. 

3) payment forwarded, when a forward succeeds. 

If we are already in the ChannelReady state and have already successfully (or unsuccessfully) forwarded the initial payment and a subsequent payment comes in. 

When it is intercepted it will be added to the queue and we will wait for one of the three triggers above to pop a payment off the queue but none of the three will happen. 

There needs to be a way to know that there's no pending payment and therefore we don't need to queue the payment but we can forward it immediately.

wvanlint:
> There needs to be a way to know that there's no pending payment and therefore we don't need to queue the payment but we can forward it immediately.

In addition, I believe we also need to verify that the queued payment is at that point sufficient to pay the fee.

I made the following changes:
- Fixed proportional fees for MPP in https://github.com/lightningdevkit/lightning-liquidity/pull/105.
- Extracted HTLC queueing logic into `PaymentQueue` in a separate commit with the original behavior.
- Generalized HTLC queueing logic:
  - The `HTLCIntercepted` event can trigger the following actions:
    - Opening of a channel if sufficient HTLCs are accumulated.
    - Forwarding of all HTLCs belonging to the corresponding payment if sufficient HTLCs are accumulated (and a previous payment failed).
    - Forwarding of the HTLC if the fee is already paid.
  - The `HTLCHandlingFailed` event will fail a pending payment, and forward a different queued payment sufficient to pay the fee if possible.
  - The `PaymentForwarded` event will forward any remaining queued payments, and disable future queueing. I was initially planning to use `skimmed_fee_msat` to verify that the skimmed fee is sufficient but that will only be available in LDK v0.0.112.

Adding additional unit test coverage to `PaymentQueue` and `OutboundJITChannelState` now.

src/lsps2/payment_queue.rs
```diff
@@ -0,0 +1,227 @@
...
+
+/// Holds payments with the corresponding HTLCs until it is possible to pay the fee.
+/// When the fee is succesfully paid with a forwarded payment, the queue should be consumed and the
+/// remaining payments forwarded.
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub(crate) struct PaymentQueue {
+	opening_fee_params: OpeningFeeParams,
+	payment_size_msat: Option<u64>,
+	opening_fee_msat: Option<u64>,
+
```
tnull: nit: Remove blank line.

src/lsps2/payment_queue.rs
```diff
@@ -0,0 +1,227 @@
+use crate::lsps2::msgs::OpeningFeeParams;
+use crate::lsps2::utils::compute_opening_fee;
+use crate::prelude::{HashMap, String, ToString, Vec};
+use alloc::collections::VecDeque;
+use lightning::ln::channelmanager::InterceptId;
+use lightning::ln::PaymentHash;
+
+/// Holds payments with the corresponding HTLCs until it is possible to pay the fee.
+/// When the fee is succesfully paid with a forwarded payment, the queue should be consumed and the
```
tnull: nit: fix typo in `successfully`.

src/lsps2/service.rs
```diff
@@ -60,118 +57,49 @@ impl From<ChannelStateError> for LightningError {
...
-				if total_expected_outbound_amount_msat >= expected_payment_size_msat
-					&& amt_to_forward_msat > 0
-				{
-					Ok(OutboundJITChannelState::PendingChannelOpen {
-						htlcs,
-						opening_fee_msat,
-						amt_to_forward_msat,
-					})
+			OutboundJITChannelState::AwaitingPayment { payment_queue } => {
+				let mut pq_guard = payment_queue.lock().unwrap();
```
tnull: nit: Elsewhere in the code we use `_lock` suffix. I have no strong opinion which one is preferable, but would prefer that we use the same suffix everywhere. (Same goes for the other `_guard`s further down.)

src/lsps2/payment_queue.rs
```diff
@@ -0,0 +1,227 @@
...
+			opening_fee_params,
+			payment_size_msat,
+			opening_fee_msat: None,
+
+			payments: HashMap::new(),
+			sufficient: VecDeque::new(),
+		}
+	}
+
+	pub(crate) fn add_htlc(
```
tnull: Hmm, not the biggest fan that suddenly all this logic is moved to the `PaymentQueue`. Is there a reasonable way to keep it in the state handling and have the `PaymentQueue` just be a data structure?

src/lsps2/service.rs
```diff
@@ -181,20 +109,46 @@ impl OutboundJITChannelState {
...
 			state => Err(ChannelStateError(format!(
 				"Channel ready received when JIT Channel was in state: {:?}",
 				state
 			))),
 		}
 	}
+
+	fn payment_forwarded(&mut self) -> Result<(Self, Vec<InterceptedHTLC>), ChannelStateError> {
+		match self {
+			OutboundJITChannelState::ChannelReady { payment_queue: Some(payment_queue) } => {
```
tnull: It's kind of confusing that this is on `ChannelState` but not a state transition really. Should we introduce a new `PaymentForwarded` state that indicates that now *everything* can just be forwarded without checking for fees?

src/lsps2/service.rs
```diff
@@ -63,11 +65,32 @@ impl From<HTLCError> for ChannelStateError {
...
+	AwaitingPayment {
+		payment_queue: Arc<Mutex<PaymentQueue>>,
+	},
+	PendingChannelOpen {
+		payment_queue: Arc<Mutex<PaymentQueue>>,
+	},
+	ChannelReady {
+		channel_id: ChannelId,
+		payment_queue: Option<Arc<Mutex<PaymentQueue>>>,
+		payment_pending: bool,
```
tnull: Yeah, let's just make "we successfully forwarded and got sufficient fee, open the floodgates" a dedicated state rather than a tracking bool. Probably, also want to rename `ChannelReady` to something like `PendingPaymentClaim` or `PendingPaymentForward`?

src/lsps2/service.rs
tnull: Still think the forwarded/failure cases should live in their separate commits.

tnull (commented):
Did a first high-level pass.

Already looks pretty good, but I'm not the biggest fan of moving all that logic into the new data structure. 

Also, we previously removed additional tracking state where possible, and should consider ways to avoid re-introducing it if possible. e.g., the intercept-id-by-channel-id tracking seems only necessary since LDK doesn't provide the `user_channel_id` in the `PaymentForwarded` event? I think we should just add that upstream rather than adding additional `HashMap`s which always also introduce additional risk of getting stale.

tnull:
Needs a rebase now that #105 landed.

src/lsps2/payment_queue.rs
wvanlint: If we move the logic out, the decision point whether a payment has accumulated sufficient HTLCs will be performed outside of the data structure. Theoretically, a max-heap of payments could be suitable where HTLCs can be queued and where we can easily verify if any payment has exceeded a certain amount.

However, I don't think we should be overly concerned with computational complexity given the bounded number of HTLCs. If we want a pure data structure, I think we can just have a `Vec<(PaymentHash, Vec<InterceptedHTLC>)>` and iterate/mutate/remove across it. I can wrap `PaymentQueue` around that, and add some utility methods (adding the HTLC and receiving the new total expected outbound amount, removing the first payment greater than a certain value). Does that sound good?

src/lsps2/service.rs
wvanlint: Sounds good! Will add this state. I'm wondering if we want a `Failed` state as well where we should fail back all incoming HTLCs.

src/lsps2/service.rs
wvanlint: There are multiple states here, separate from a successful forward. The original states are as follows and can be renamed:
- `ChannelReady { payment_queue: None, .. }`: The payment has been successfully forwarded, will be renamed to `PaymentForwarded`.
- `ChannelReady { payment_pending: true, payment_queue: Some(_), .. }`:  A payment is pending, can be renamed to `PendingPaymentForward`.
- `ChannelReady { payment_pending: false, payment_queue: Some(_), .. }`: Initial payment(s) failed (such as prepayment probes), and a subsequent payment has not arrived yet. It is similar to `AwaitingPayment` but the channel is already open so not sure what the best way to rename this is. Maybe rename the original `AwaitingPayment` to `AwaitingInitialPayment` and this one to `AwaitingSecondaryPayment`? But that does not feel ideal.

src/lsps2/payment_queue.rs
tnull: Yes, that sounds good to me. I agree that iterating a `Vec` should be alright here.

src/lsps2/service.rs
tnull: I'm not sure we need to add such a state as we probably want to drop the state entirely once we're sure that it failed irrevocably? So rather than transitioning to that state, we should just signal to fail all HTLCs, close the channel and delete the state?

src/lsps2/service.rs
tnull: I agree we should add more states here, but I think we don't necessarily have to split between first/second HTLC, we can just stay in `PendingPaymentForward` until we have sufficient parts. I'd however agree that renaming `AwaitingPayment` would make sense now. Maybe `Pending(Initial)HTLCIntercept`?

src/lsps2/payment_queue.rs
wvanlint: Done.

src/lsps2/payment_queue.rs
wvanlint: Done.

src/lsps2/service.rs
wvanlint: Moved to `_lock` suffix.

src/lsps2/payment_queue.rs
wvanlint: Changed `PaymentQueue` to be a pure data structure with a specific API.

src/lsps2/service.rs
wvanlint: Yeah, sounds perfect. We only fail irrevocably in this case? https://github.com/lightningdevkit/lightning-liquidity/blob/99b2960ff3eaac0cca7c1c045a839c7562cdc07d/src/lsps2/service.rs#L488-L497 It seems like this can only happen for the first HTLC.

src/lsps2/service.rs
wvanlint: Added additional states. I think there is a distinction between the interception of the initial payment that triggers the opening of the channel, and the interception of a subsequent payment if the initial payment fails (`PendingInitialPayment` and `PendingPayment` respectively). Hopefully the code illustrates the cases but let me know if I missed anything.

src/lsps2/service.rs
wvanlint: Separated the commits.

wvanlint:
> Also, we previously removed additional tracking state where possible, and should consider ways to avoid re-introducing it if possible. e.g., the intercept-id-by-channel-id tracking seems only necessary since LDK doesn't provide the user_channel_id in the PaymentForwarded event? I think we should just add that upstream rather than adding additional HashMaps which always also introduce additional risk of getting stale.

I agree that this can probably be upstreamed in a subsequent version of LDK. In the meantime, I think it would be worthwhile to add this and revert it after the upgrade.

src/lsps2/payment_queue.rs
```diff
@@ -0,0 +1,113 @@
...
+impl PaymentQueue {
+	pub(crate) fn new() -> PaymentQueue {
+		PaymentQueue { payments: Vec::new() }
+	}
+
+	pub(crate) fn add_htlc(&mut self, htlc: InterceptedHTLC) -> (u64, usize) {
+		for (payment_hash, htlcs) in self.payments.iter_mut() {
+			if payment_hash == &htlc.payment_hash {
+				htlcs.push(htlc);
+				return (htlcs.iter().map(|htlc| htlc.expected_outbound_amount_msat).sum(), htlcs.len());
```
tnull: nit: Let's move the iterator here out to a `total_expected_outbound_amount_msat` variable or similar.

src/lsps2/payment_queue.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl PaymentQueue {
+	pub(crate) fn new() -> PaymentQueue {
+		PaymentQueue { payments: Vec::new() }
+	}
+
+	pub(crate) fn add_htlc(&mut self, htlc: InterceptedHTLC) -> (u64, usize) {
+		for (payment_hash, htlcs) in self.payments.iter_mut() {
+			if payment_hash == &htlc.payment_hash {
+				htlcs.push(htlc);
```
tnull: Should we `debug_assert!` that the HTLC with given intercept id isn't already present here? Similarly, could `debug_assert` that all iterated `InterceptedHTLC::payment_hash`es match the 'key' payment hash.

src/lsps2/payment_queue.rs
```diff
@@ -0,0 +1,113 @@
...
+	pub(crate) payment_hash: PaymentHash,
+}
+
+impl PaymentQueue {
+	pub(crate) fn new() -> PaymentQueue {
+		PaymentQueue { payments: Vec::new() }
+	}
+
+	pub(crate) fn add_htlc(&mut self, htlc: InterceptedHTLC) -> (u64, usize) {
+		for (payment_hash, htlcs) in self.payments.iter_mut() {
```
tnull: A pattern like
```rust
if let Some(..) = self.payments.iter_mut().find(|(p, _)| p == htlc.payment_hash) { 
    .. 
} else { 
    ...
}
```

would IMO improve readability and let us avoid the (error prone) early-return pattern here.

src/lsps2/service.rs
```diff
@@ -46,10 +49,16 @@ pub struct LSPS2ServiceConfig {
 	pub promise_secret: [u8; 32],
 }
 
-#[derive(Copy, Clone, PartialEq, Eq, Debug)]
-struct InterceptedHTLC {
-	intercept_id: InterceptId,
-	expected_outbound_amount_msat: u64,
+#[derive(Clone, Debug, PartialEq)]
+struct OpenChannelParams {
```
tnull, line 53: Even though not exposed, these structs could use a one-liner describing what they represent, or at least some more self-explanatory names. E.g., I can make different assumptions what a `FeePayment` could be, but I'll hardly know without reading all the connected code for context.

src/lsps2/service.rs
```diff
@@ -152,21 +154,22 @@ impl OutboundJITChannelState {
 				if total_expected_outbound_amount_msat >= expected_payment_size_msat
 					&& amt_to_forward_msat > 0
 				{
-					Ok(OutboundJITChannelState::PendingChannelOpen {
-						htlcs,
+					Ok((OutboundJITChannelState::PendingChannelOpen {
+						payment_queue: payment_queue.clone(),
```
tnull: nit: Kind of personal preference, but would prefer to use `Arc::clone(&payment_queue)` to clearly disambiguate copying references from deep copies.

src/lsps2/service.rs
```diff
@@ -60,87 +69,80 @@ impl From<ChannelStateError> for LightningError {
...
 enum OutboundJITChannelState {
-	AwaitingPayment {
-		min_fee_msat: u64,
-		proportional_fee: u32,
-		min_payment_size_msat: u64,
-		max_payment_size_msat: u64,
-		htlcs: Vec<InterceptedHTLC>,
+	PendingInitialPayment {
 		payment_size_msat: Option<u64>,
+		opening_fee_params: OpeningFeeParams,
```
tnull: Can we keep these by reference to avoid having to clone them below?

src/lsps2/service.rs
```diff
@@ -542,13 +547,18 @@ where
 	pub fn channel_ready(
 		&self, user_channel_id: u128, channel_id: &ChannelId, counterparty_node_id: &PublicKey,
 	) -> Result<(), APIError> {
+		{
+			let mut peer_by_channel_id = self.peer_by_channel_id.write().unwrap();
+			peer_by_channel_id.insert(*channel_id, *counterparty_node_id);
```
tnull, line 862: I agree that we probably as of now need to keep this state. However, do we have a plan when to clean it up? Since we're going to persist a lot of these things going forward, can we remove these mappings again as soon as we drop the peer. Thinking of it, can we drop the peer state once we either fulfilled all pending requests or failed all HTLC of all channels? Maybe all this cleanup would be a reasonable follow-up PR though?

src/lsps2/service.rs
```diff
@@ -155,20 +171,20 @@ impl OutboundJITChannelState {
 					&& amt_to_forward_msat > 0
 				{
 					Ok((OutboundJITChannelState::PendingChannelOpen {
```
tnull: nit: While we're here, can we move these returned values in to variables to declutter the `Ok(..)` a bit? Same below.

src/lsps2/service.rs
```diff
@@ -538,6 +610,71 @@ where
 		Ok(())
 	}
 
+	/// Forward [`Event::PaymentForwarded`] event parameter into this function.
+	///
+	/// Will register the forwarded payment as having paid the JIT channel fee, and
+	/// forward any held and future HTLCs for the SCID of the initial invoice.
+	/// In the future, this will verify skimmed_fee_msat.
```
tnull: nit: tick `skimmed_fee_msat`, or, even better, tick and link to it.

src/lsps2/service.rs
```diff
@@ -538,6 +610,71 @@ where
...
+	///
+	/// [`Event::PaymentForwarded`]: lightning::events::Event::PaymentForwarded
+	pub fn payment_forwarded(&self, next_channel_id: Option<ChannelId>) -> Result<(), APIError> {
```
tnull: We could consider requiring the `ChannelId` in the API here, since it will be present in LDK >= v0.0.107, and `lightning-liquidity` was introduced well later than that. If we want to keep it an option, we should def. return an error and document that it's technically required.

src/lsps2/payment_queue.rs
```diff
@@ -26,7 +26,10 @@ impl PaymentQueue {
 		for (payment_hash, htlcs) in self.payments.iter_mut() {
 			if payment_hash == &htlc.payment_hash {
 				htlcs.push(htlc);
-				return (htlcs.iter().map(|htlc| htlc.expected_outbound_amount_msat).sum(), htlcs.len());
+				return (
+					htlcs.iter().map(|htlc| htlc.expected_outbound_amount_msat).sum(),
```
tnull: nit: This formatting change probably belongs in the first commit.

src/lsps2/service.rs
```diff
@@ -125,7 +132,8 @@ impl OutboundJITChannelState {
 				opening_fee_params,
 				payment_queue,
 			} => {
-				let (total_expected_outbound_amount_msat, num_htlcs) = payment_queue.lock().unwrap().add_htlc(htlc);
+				let (total_expected_outbound_amount_msat, num_htlcs) =
+					payment_queue.lock().unwrap().add_htlc(htlc);
```
tnull, line 137: nit: Same here, reformatting needs to happen in one of the prior commits. Goes for most of the changes below I think.

tnull (commented):
This is much cleaner, thanks!

Looks great, mod few comments. It also needs a minor rebase and the individual commits should probably be reformatted to avoid formatting changes happening in later commits.

I'm still not convinced we need a separate `PendingPayment` state. Isn't it identical to `PendingPaymentForward` and any difference could be expressed via the returned `ForwardAction` being `None`/`Some`? Could you otherwise explain how they differ?

src/lsps2/payment_queue.rs
wvanlint: Done.

src/lsps2/payment_queue.rs
wvanlint: Added `debug_assert!` cases for both statements.

src/lsps2/payment_queue.rs
wvanlint: Done.

src/lsps2/service.rs
wvanlint: Added comments on the structs here.

src/lsps2/service.rs
wvanlint: Done.

src/lsps2/service.rs
wvanlint: Hmm, it turned out a bit difficult due to lifetimes and the structs are small, but let me know if I missed an approach here.

src/lsps2/service.rs
wvanlint: >  Thinking of it, can we drop the peer state once we either fulfilled all pending requests or failed all HTLC of all channels? Maybe all this cleanup would be a reasonable follow-up PR though?

Yes I believe so. This applies to all the original state as well - doing this in a follow-up sounds perfect.

src/lsps2/service.rs
wvanlint: Reorganized these into variables.

src/lsps2/service.rs
wvanlint: Applied ticks, but linking is not possible until we upgrade I believe.

src/lsps2/service.rs
wvanlint: Returned an `APIError` when `ChannelId` is not present. Mainly took this approach so users don't have to duplicate effort unwrapping this.

src/lsps2/payment_queue.rs
wvanlint: Fixed.

src/lsps2/service.rs
wvanlint: Fixed.

wvanlint:
> I'm still not convinced we need a separate PendingPayment state. Isn't it identical to PendingPaymentForward and any difference could be expressed via the returned ForwardAction being None/Some? Could you otherwise explain how they differ?

In both states, the channel is open and we have not received a successful confirmation of a payment forward yet.
The main difference is that for `PendingPaymentForward`, we have a payment forward in progress, while for `PendingPayment`, any previous payments have failed.

If a different, complete payment comes in, we can't forward it while being in `PendingPaymentForward`. We don't know whether the pending forward will succeed or fail, so we don't know whether to skim a fee or not.

If such a payment comes in while being in `PendingPayment`, we know no other payments are pending and can skim the fee and forward the payment.

src/lsps2/service.rs
tnull: I mean the easiest way if we don't want to deal with lifetimes would probably be to us an `Rc`. But you're right, the struct is not massive, should be ~sub 100 bytes. Still handing around a reference would be preferable.

src/lsps2/service.rs
tnull: Okay, although providing an API with two options and have one of them always fail is a bit weird, tbh. Might be preferable to let the user do the unwrapping instead of letting them discover on runtime that something is wrong.

tnull:
> > I'm still not convinced we need a separate PendingPayment state. Isn't it identical to PendingPaymentForward and any difference could be expressed via the returned ForwardAction being None/Some? Could you otherwise explain how they differ?
> 
> In both states, the channel is open and we have not received a successful confirmation of a payment forward yet. The main difference is that for `PendingPaymentForward`, we have a payment forward in progress, while for `PendingPayment`, any previous payments have failed.
> 
> If a different, complete payment comes in, we can't forward it while being in `PendingPaymentForward`. We don't know whether the pending forward will succeed or fail, so we don't know whether to skim a fee or not.
> 
> If such a payment comes in while being in `PendingPayment`, we know no other payments are pending and can skim the fee and forward the payment.

Aah, okay that makes sense. Do you mind adding comments to the variants in `OutboundJITChannelState` to describe their differences and any assumptions we make around them? That would help us going forward to be aware of the rationale of the chosen design without coming back here to investigate.

src/lsps2/service.rs
wvanlint: Moved these into `OutboundJITChannel` so they can be passed by reference into the `OutboundJITChannelState` method.

src/lsps2/service.rs
wvanlint: Changed this to requiring `ChannelId`. I was thinking more towards at some point taking in the `PaymentForwarded` event, but this API makes sense right now.

wvanlint:
Added documentation around the different states on `OutboundJITChannelState`.

src/sync/nostd_sync.rs
```diff
@@ -10,6 +11,13 @@ pub struct Mutex<T: ?Sized> {
 	inner: RefCell<T>,
 }
 
+impl<T: fmt::Debug> fmt::Debug for Mutex<T> {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		let t = self.lock().unwrap();
```
tnull, line 16: Usually, you'd want to use `try_lock` here, but, given they are identical in this context, this should be fine.

tnull (approved):
LGTM, thanks!

PR #105, opened 2024-02-19
Fixes proportional fees for MPP

wvanlint:
Proportional fee attribution across HTLCs contained a rounding error, dividing two integers whose value would always be smaller than 1. https://github.com/lightningdevkit/lightning-liquidity/blob/71b9dd2a8df3c5d0bae43dd442c7c225cf0763f4/src/lsps2/service.rs#L775-L776 This resulted in the fee being solely assigned to the final HTLC. https://github.com/lightningdevkit/lightning-liquidity/blob/71b9dd2a8df3c5d0bae43dd442c7c225cf0763f4/src/lsps2/service.rs#L782-L784 This PR applies a fix, adds a unit test, and adjusts the proptest for coverage.

The method now also takes in the fee as opposed to the amount to forward. This was done for clarity and for the future case where an initial payment determines the channel parameters and fails, but a subsequent payment of a different size succeeds and pays for the fee.

### DIFF
--- a/src/lsps2/service.rs
+++ b/src/lsps2/service.rs
@@ -77,7 +77,7 @@ enum OutboundJITChannelState {
 	},
 	ChannelReady {
 		htlcs: Vec<InterceptedHTLC>,
-		amt_to_forward_msat: u64,
+		opening_fee_msat: u64,
 	},
 }
 
@@ -183,10 +183,10 @@ impl OutboundJITChannelState {
 
 	fn channel_ready(&self) -> Result<Self, ChannelStateError> {
 		match self {
-			OutboundJITChannelState::PendingChannelOpen { htlcs, amt_to_forward_msat, .. } => {
+			OutboundJITChannelState::PendingChannelOpen { htlcs, opening_fee_msat, .. } => {
 				Ok(OutboundJITChannelState::ChannelReady {
 					htlcs: htlcs.clone(),
-					amt_to_forward_msat: *amt_to_forward_msat,
+					opening_fee_msat: *opening_fee_msat,
 				})
 			}
 			state => Err(ChannelStateError(format!(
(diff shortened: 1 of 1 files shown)


src/lsps2/service.rs
```diff
@@ -759,40 +757,37 @@ where
 }
 
 fn calculate_amount_to_forward_per_htlc(
-	htlcs: &[InterceptedHTLC], total_amt_to_forward_msat: u64,
+	htlcs: &[InterceptedHTLC], total_fee_msat: u64,
 ) -> Vec<(InterceptId, u64)> {
 	// TODO: we should eventually make sure the HTLCs are all above ChannelDetails::next_outbound_minimum_msat
 	let total_received_msat: u64 =
```
tnull: nit: While we're here we could improve the naming of this variable. `total_received_msat` is a bit confusing as it's not actually the (inbound) amount we received, but the sum of all the expected *outbound* amounts of all HTLCs.

src/lsps2/service.rs
```diff
@@ -759,40 +757,37 @@ where
...
 		htlcs.iter().map(|htlc| htlc.expected_outbound_amount_msat).sum();
+	if total_fee_msat > total_received_msat {
+		return Vec::new();
```
tnull, line 767: Could we add a `debug_assert` here? I think this should never get hit and we should make sure we catch it early in tests.

src/lsps2/service.rs
```diff
@@ -843,16 +838,42 @@ mod tests {
...
-				let fair_share_2 = ((o_2 as f32 / total_received_msat as f32) * result_sum as f32).max(o_2 as f32);
+				let fair_share_2 = (o_2 as f32 / total_received_msat as f32) * result_sum as f32;
 				assert!(result[2].1 as f32 <= fair_share_2 + five_pct);
 			}
-
 		}
 	}
+
+	#[test]
+	fn test_calculate_amount_to_forward() {
```
tnull, line 855: Not sure what additional coverage this test adds on top of the proptest though?

tnull (commented):
Ah, good catch. Basically LGTM, just a few questions.

src/lsps2/service.rs
wvanlint: Renamed to `total_expected_outbound_msat`.

src/lsps2/service.rs
wvanlint: Done.

src/lsps2/service.rs
wvanlint: The previous version of the proptest had a bug, so I wanted to at least have one unit test with specific values that can be verified exactly without margins.

tnull: Yeah, I guess, no that much overhead anyways.

tnull (approved):

PR #116, opened 2024-02-29
Fix deadlock on deserialization failure

tnull:
Previously, we introduced disconnecting a counterparty who sends us bogus messages that we're unable to parse. However, before we disconnect we also send out a last `LSPSMessage::Invalid` in the hopes that the
counterparty understands this.

However, when we added this logic we unfortunately overlooked that we lock the `request_id_to_method_map` `Mutex` for parsing the message, but also try to lock when the `PeerHandler` calls `get_and_clear_pending_msg`.

Here, we avoid the resulting deadlock by dropping the lock as soon as it's not required anymore after parsing.

In a second commit, we avoid unnecessary locking of `request_ids_and_methods_map` in `get_and_clear_pending_msg` and  remove a potentially dangerous (if we ever were to fail serialization for some reason) `unwrap`.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -481,22 +481,23 @@ where
 		&self, msg: Self::CustomMessage, sender_node_id: &PublicKey,
 	) -> Result<(), lightning::ln::msgs::LightningError> {
 		let message = {
-			let mut request_id_to_method_map = self.request_id_to_method_map.lock().unwrap();
-			LSPSMessage::from_str_with_id_map(&msg.payload, &mut request_id_to_method_map).map_err(
-				|_| {
-					let error = ResponseError {
-						code: JSONRPC_INVALID_MESSAGE_ERROR_CODE,
-						message: JSONRPC_INVALID_MESSAGE_ERROR_MESSAGE.to_string(),
-						data: None,
-					};
+			{
+				let mut request_id_to_method_map = self.request_id_to_method_map.lock().unwrap();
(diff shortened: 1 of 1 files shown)


src/manager.rs
```diff
@@ -480,41 +483,62 @@ where
...
 	) -> Result<(), lightning::ln::msgs::LightningError> {
+		{
+			if self.ignored_peers.read().unwrap().contains(&sender_node_id) {
+				let err = format!("Ignoring message from peer {}.", sender_node_id);
+				return Err(LightningError {
+					err,
+					action: ErrorAction::IgnoreAndLog(Level::Trace),
+				});
+			}
+		}
```
johncantrell97, line 494: Will this respond to peer with an error or just log it on our end? We've run into issues where we just silently fail/ignore requests (when user hits our rate limits) and the wallet user experience is odd because it just hangs.  This could introduce that experience?

johncantrell97 (approved):

src/manager.rs
tnull: Yes, but in this case it's really intended as it's a DoS protection if the user sends us bogus data. The alternative is disconnecting them which I'd even prefer, but we don't know if the user has any  channels with us that still need to be kept operational. While this just adds the user to the ignorelist until restart, we def. want to remove them after a while in the future, especially once we start persisting things, as tracked here: https://github.com/lightningdevkit/lightning-liquidity/issues/117

PR #118, opened 2024-03-02
Add integration test utils, first LSPS2 integration test

tnull:
Here, we add some integration test utils and add an initial integration test that covers the LSPS2 invoice generation flow up to creating a new invoice. In the future we'll extend that to create a full-cycle test.

However, as this is already quite a bit of code, extending it to a full-cycle test will happen in a follow-up PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,7 @@ jobs:
         if: matrix.msrv
         run: |
           cargo update -p proptest --precise "1.2.0" --verbose # proptest 1.3.0 requires rustc 1.64.0
+          cargo update -p regex --precise "1.9.6" --verbose # regex 1.10.0 requires rustc 1.65.0.
       - name: Cargo check
         run: cargo check --release
       - name: Check documentation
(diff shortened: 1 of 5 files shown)


tnull:
Rebased for LDK v0.0.123.

@johncantrell97 Would be great to get some intial review on this.

johncantrell97:
This looks fine to me, I don't really have any comments to add.  It seems like most of the stuff in common is directly from ldk, right?

Nice that you're able to use the library in this modular way to pipe the request/responses in and out without any changes.

johncantrell97 (approved):

tnull:
> This looks fine to me, I don't really have any comments to add. It seems like most of the stuff in common is directly from ldk, right?

Yes, mostly stolen from the `BackgroundProcessor` tests, but deleted a lot of unnecessary code and modified some other parts.

> Nice that you're able to use the library in this modular way to pipe the request/responses in and out without any changes.

Yep!

tnull:
Fixed MSRV CI, going ahead and merge this for now. We'll interate on the test (utils) in the future anyways.

PR #125, opened 2024-03-25
Fix more potential deadlock scenarios

tnull:
While we previously already fixed some potential deadlock issues, some more remained in the LSPS1/LSPS2 implementations due to holding the peer locks when enqueuing request/response messages as the latter would trigger `PeerManager::process_events` could result in circular waits. Here, we make sure to always drop the locks first, before enqueuing request/response messages.

### DIFF
--- a/src/lsps1/client.rs
+++ b/src/lsps1/client.rs
@@ -243,20 +243,21 @@ where
 	pub fn send_get_info_request(&self, counterparty_node_id: PublicKey, user_channel_id: u128) {
 		let channel = InboundCRChannel::new(user_channel_id);
 
-		let mut outer_state_lock = self.per_peer_state.write().unwrap();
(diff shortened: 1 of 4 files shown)


wvanlint (approved):
LGTM

I was wondering whether we can extract enqueueing the message and not repeat it, but that might only be easy for the `handle_*` methods.

tnull:
> LGTM
> 
> I was wondering whether we can extract enqueueing the message and not repeat it, but that might only be easy for the `handle_*` methods.

Yeah, we could eventually consider a larger refactor to some structure that would prohibit us from ever holding the lock while enqueueing. This might be a good idea, but if we do that however, I'd like to share it across LSPS1/2 at least, which might require some more prefactors to make them more uniform. 

Landing this for now.

PR #137, opened 2024-05-17
LSPS2: ability to signal channel open failure

### LINKED ISSUE #3479: LSPS2: Fail intercepted htlcs if lsp open channel fails for whatever reason
There's a scenario where we emit `OpenChannel` event to the LSP to handle.  From there `liquidity` is waiting for them to feed us a `ChannelReady` event but if they refuse or fail to open the channel then this HTLC is just stuck forever.

We should expose a method the LSP can use to tell us to go ahead and fail all the pending payments backwards.

johncantrell97:
Closes https://github.com/lightningdevkit/rust-lightning/issues/3479

There are a number of reasons a channel open might fail or be abandoned by the LSP. This adds a way for the LSP to signal back to the library that this has happened.

All pending payments are failed backwards and the channel state is reset to PendingInitialPayment as per the spec. This allows the payer to try again.

### DIFF
--- a/src/lsps2/payment_queue.rs
+++ b/src/lsps2/payment_queue.rs
@@ -49,8 +49,8 @@ impl PaymentQueue {
 		position.map(|position| self.payments.remove(position))
 	}
 
-	pub(crate) fn clear(&mut self) -> Vec<InterceptedHTLC> {
-		self.payments.drain(..).map(|(_k, v)| v).flatten().collect()
+	pub(crate) fn clear(&mut self) -> Vec<(PaymentHash, Vec<InterceptedHTLC>)> {
+		self.payments.drain(..).collect()
 	}
 }
 
@@ -109,11 +109,14 @@ mod tests {
 		);
 		assert_eq!(
 			payment_queue.clear(),
-			vec![InterceptedHTLC {
-				intercept_id: InterceptId([1; 32]),
-				expected_outbound_amount_msat: 300_000_000,
-				payment_hash: PaymentHash([101; 32]),
-			}]
+			vec![(
+				PaymentHash([101; 32]),
+				vec![InterceptedHTLC {
+					intercept_id: InterceptId([1; 32]),
+					expected_outbound_amount_msat: 300_000_000,
+					payment_hash: PaymentHash([101; 32]),
+				}]
+			)]
 		);
 	}
 }
--- a/src/lsps2/service.rs
+++ b/src/lsps2/service.rs
@@ -19,7 +19,7 @@ use crate::prelude::{HashMap, String, ToString, Vec};
 use crate::sync::{Arc, Mutex, RwLock};
 
 use lightning::events::HTLCDestination;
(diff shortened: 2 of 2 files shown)


src/lsps2/service.rs
```diff
@@ -898,6 +904,79 @@ where
 		Ok(())
 	}
 
+	/// Used by LSP to fail intercepted htlcs backwards when the channel open fails for any reason.
+	///
+	/// Should be called in response to receiving a [`LSPS2ServiceEvent::OpenChannel`] event.
+	///
+	/// The JIT channel state is reset such that the payer can attempt payment again.
```
tnull, line 911: I wonder if we also expose a "abandon" variant that *doesn't* reset, but just prunes all state related to the channel open?

src/lsps2/service.rs
```diff
@@ -898,6 +904,79 @@ where
 		Ok(())
 	}
 
+	/// Used by LSP to fail intercepted htlcs backwards when the channel open fails for any reason.
+	///
+	/// Should be called in response to receiving a [`LSPS2ServiceEvent::OpenChannel`] event.
```
tnull, line 909: This comment is a bit misleading. It should be called after we get the event, but of course only if the channel open failed?

src/lsps2/payment_queue.rs
```diff
@@ -49,8 +49,8 @@ impl PaymentQueue {
 		position.map(|position| self.payments.remove(position))
 	}
 
-	pub(crate) fn clear(&mut self) -> Vec<InterceptedHTLC> {
-		self.payments.drain(..).map(|(_k, v)| v).flatten().collect()
+	pub(crate) fn clear(&mut self) -> Vec<(PaymentHash, Vec<InterceptedHTLC>)> {
```
tnull, line 52: nit: Not sure this change is worth as it makes things more complex and you only make use of it in one place where you then have to `map` to get rid of the `InterceptedHTLC`s. Why not just use `InterceptedHTLC::payment_hash` instead? If you're worried about duplicates, could just use `dedup` or collect a `HashSet`?

tnull (commented):
Excuse the delay here!

LGTM, mod some minor questions/nits.

tnull:
Any intentions to pick this back up?

johncantrell97:
> Any intentions to pick this back up?

Yeah probably, though I'm not really sure what the status of the whole LSPS project is? Wasn't the latest to transition to a liquidity ads based solution?

tnull:
> Yeah probably, though I'm not really sure what the status of the whole LSPS project is? Wasn't the latest to transition to a liquidity ads based solution?

Well, we still intend to finish LSPS1/LSPS2 work, even though future iterations will be based on Liquidity Ads. 

Cool to hear you might be interested in getting the open PRs landed. Note however that we're finally about to move `lightning-liquidity` into `rust-lightning`, so you might need to reopen them there once that is done.

tnull:

PR #139, opened 2024-07-08
Remove max_level_trace feature on lightning package

wvanlint:
Cargo resolves dependencies with the union of all features enabled on them. Setting the max_level_trace feature stops any consumers from emitting gossip logs.


tnull (approved):
Makes sense.

